PR #5, opened 2016-01-07
Validate all links are present

atomaka:
Since I began using the companion before the 1.0.0 release, I had an issue where my I did not have a `dhparam.pem` file.  After adding one, appropriate links were not created by the application because linking only occurred after a successful certificate renewal.  While touching this, it was also clear that users could run into an issue where a key or certificate link was removed accidentally and then never recreated.

This pull request will rebuild links if they are missing or if the simp_le client creates or updates a certificate.

### DIFF
--- a/app/letsencrypt_service
+++ b/app/letsencrypt_service
@@ -7,11 +7,19 @@ acme_ca_uri="${ACME_CA_URI:-https://acme-v01.api.letsencrypt.org/directory}"
 
 source /app/functions.lib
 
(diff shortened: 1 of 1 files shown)


atomaka:
This is actually an issue with links to the key and certificates as well.  I'm updating to account for that.  Will rebase when complete.

atomaka:
Should be done messing around with with this now, @JrCs.

Issue #12, opened 2016-01-24
Usage with separate nginx and docker-gen container


konstantinj:
Hi,

thanks for this awesome image - exactly what I was looking for. Unfortunately it's not working with my configuration. I'm using docker-compose and I have a separate nginx container and docker-gen in its own container so I can modify the nginx config template.

I'm getting this error on start:
> ,sh: 1: /usr/local/bin/docker-gen: not found

How is this image supposed to work?

JrCs:
My container call the nginx proxy container to recreate the nginx configuration with docker-gen.
Why don't you use the original nginx-proxy container with docker-gen in it ?
If you need to change the template that is used to create the nginx configuration you can mount bind the nginx.tpl with your configuration.

konstantinj:
The default nginx-proxy container is using the /app/nginx.tmpl file which I can't overwrite with a mount. Also by using nginx separately I can use a newer version and the docker socket won't be exposed publicly.

bobmaerten:
Exactly my setup, @konstantinj. I'll try installing/using the @JrCs image to get it running, and I'll report back here then.

JrCs:
To use my container we need to know the container id of the docker-gen container.
@konstantinj is your docker-gen container do an nginx reload ? if yes how ?

konstantinj:
In the readme from https://github.com/jwilder/nginx-proxy it says that the nginx container must be named "nginx". So I guess that's how docker-gen can reload nginx. So it should work the same with your container I guess when the container name is hard "docker-gen".

bobmaerten:
Well, no luck with containerized `docker-gen`, so I switch back on default nginx-proxy for now.

JrCs:
@konstantinj no the docker-gen command of the nginx-proxy call directly the nginx reload command (they are in the same container)

konstantinj:
Yes, when using nginx-proxy image. But in my case, when using docker-gen image and nginx image separately, docker-gen is reloading nginx in the other container.

JrCs:
How the docker-gen image reload the nginx in the other container ?

bobmaerten:
@JrCs `docker-gen` tool has a `-notify-sighup <container_name>` flag to do some sort of `kill -HUP` to the container.

falcon15500:
[...]

PR #20, opened 2016-02-04
Support for separate nginx and docker-gen containers

nilroy:
This PR addresses the case where docker-gen and nginx are in separate containers

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ FROM alpine:latest
 MAINTAINER Yves Blusseau <90z7oey02@sneakemail.com> (@blusseau)
 
 ENV DEBUG=false              \
-    DOCKER_GEN_VERSION=0.5.0 \
+    DOCKER_GEN_VERSION=0.4.3 \
     DOCKER_HOST=unix:///var/run/docker.sock
 
 RUN apk --update add bash curl ca-certificates tar procps jq && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ FROM alpine:latest
 MAINTAINER Yves Blusseau <90z7oey02@sneakemail.com> (@blusseau)
 
 ENV DEBUG=false              \
-    DOCKER_GEN_VERSION=0.4.3 \
+    DOCKER_GEN_VERSION=0.5.0 \
     DOCKER_HOST=unix:///var/run/docker.sock
 
 RUN apk --update add bash curl ca-certificates tar procps jq && \
--- a/README.md
+++ b/README.md
@@ -43,6 +43,23 @@ Then start any containers you want to proxied with a env var `VIRTUAL_HOST=subdo
 
 The containers being proxied must [expose](https://docs.docker.com/reference/run/#expose-incoming-ports) the port to be proxied, either by using the `EXPOSE` directive in their `Dockerfile` or by using the `--expose` flag to `docker run` or `docker create`. See [nginx-proxy](https://github.com/jwilder/nginx-proxy) for more informations. To generate automatically Let's Encrypt certificates see next section.
 
+#### Separate Containers
+nginx-proxy can also be run as two separate containers using the [jwilder/docker-gen](https://github.com/jwilder/docker-gen)
+image and the official [nginx](https://hub.docker.com/_/nginx/) image.
+ 
+* First start nginx and docker-gen containers separately following the instructions in [separate-container-install](https://github.com/jwilder/docker-gen#separate-container-install)
+
+* Second start this container:
+```bash
+$ docker run -d \
+    -v /path/to/certs:/etc/nginx/certs:rw \
+    --volumes-from nginx-proxy \
+    -v /var/run/docker.sock:/var/run/docker.sock:ro \
+    -v /etc/docker-gen/templates:/etc/docker-gen/templates \
+    jrcs/letsencrypt-nginx-proxy-companion
+```
+Then start any containers to be proxied as described previously.
+
 #### Let's Encrypt
 
 To use the Let's Encrypt service to automatically create a valid certificate for virtual host(s).
--- a/README.md
+++ b/README.md
@@ -55,7 +55,8 @@ $ docker run -d \
     -v /path/to/certs:/etc/nginx/certs:rw \
     --volumes-from nginx-proxy \
     -v /var/run/docker.sock:/var/run/docker.sock:ro \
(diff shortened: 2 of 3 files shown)


nilroy:
@JrCs --> We are using separate docker-gen and nginx containers for some security reasons. This PR addresses our case and at the same time it works with the bundled nginx-proxy also.

JrCs:
Thanks nilroy, i will merge it as soon as possible

JrCs:
Will close #12

falcon15500:
So the LE container has /usr/local/bin/docker-gen (I wasn't aware), then this is even easier and better than my currently (mostly) working implementation.  Get the LE container to update the default.conf via the same template and just restart nginx.  Very simple and great!  Am testing it now locally (mainly because I can't wait).  Thanks!

falcon15500:
Actually...  I note that the nginx.tmpl file that comes with docker-gen 0.5.0 is very basic and doesn't cater at all to SSL.  I am currently using my own version of that file which is an amalgamation of the one from the current nginx-proxy and the newer network specific parts of the docker-gen 0.5.0 template...

What are you guys using for this?

JrCs:
@nilroy the LE container have the docker-gen command so we doesn't need another docker-gen container, but only the right nginx.tmpl file

nilroy:
@JrCs Yes you are right...I am currently testing the stuf without the docker-gen container. However I could not use docker-gen 0.5.0 as it does not exited  out after generating the nginx files...Probably there are some options that I am not aware of. Probably the readme needs to be changed a bit. Also do you have a plan to keep the nginx.tmpl file as part of this repo?

nilroy:
@falcon15500  I am using nginx.tmpl from nginx-proxy stuff. Also have you encountered rate limits for letsencrypt?

nilroy:
@JrCs I tested. It works without docker-gen container only when we use SSL and there is nothing wrong in it. The LE container is designed for SSL. Now I have adifferent scenario. In my environment we need to use mixed type of containers where some will use SSL and some not. I would still use the setup that contain both docker-gen and the LE container. I would leave a note in the README about it. So that people can choose whatever that fits them. What do you say?

falcon15500:
I have the mixed scenario working using the files from this pull request along with my own modified template (and docker-gen 0.5.0). This container handles the ssl stuff and the docker-gen container handles the rest.

They both reference the same template.

In regards to the issue with docker-gen 0.5.0 not exiting, I encountered this only when calling the docker-gen command remotely from the LE container (via that api exec call). However it seems to be working okay by itself.

I did raise it as an issue in the docker-gen github.

nilroy:
@falcon15500 I have encountered the issue(docker-gen not exiting)  even If ran it on the docker-gen container. But let them solve it anyways. In my environment also we are using the same nginx.tmpl file for docker-gen and LE container. Now I have another problem. I have encountered rate limit from ACME.

JrCs:
The problem is nginx.tmpl. I don’t want to have a specific version in my container. 
This container was solely created to generate letsencrypt certificates, not to manage nginx proxy configuration(s).

nilroy:
@JrCs then lets not maintain nginx.tmpl here. But can we merge this PR?

JrCs:
@nilroy yes but i need some time to review/test it.

nilroy:
@JrCs No hurry. :+1:

falcon15500:
Problem I currently have, is that the version of docker-gen included within this LE container isn't new enough to accept the docker networking changes with the newer nginx.tmpl file I am using...

Any chance you can upgrade the docker-gen binary?

nilroy:
I can test if that works.

falcon15500:
Using docker-gen 0.5.0 definitely doesn't work due to the already reported "non-exit" bug.  I tried the binary from 0.4.3 as well, but it's not updated with the new networking stuff so I can't use it.

Hopefully the 0.5.0 issue will be sorted soon, as that is all that I need to have this all working seemlessly.

nilroy:
@falcon15500. I tested docker-gen 0.5.0 in LE container and it works. It is exiting properly while it is run locally inside the LE container.

nilroy:
@JrCs I am updating this PR

falcon15500:
@nilroy Definitely doesn't exit for me, even when run manually. I need to CTRL C to kill it. Tried it this morning.

nilroy:
@falcon15500  Sorry... Its not exiting. I was wrongly testing. Its generated the configuration file but did not reloaded nginx. My bad

falcon15500:
Okay, jwilder has just put through a commit that fixes the "not exiting" problem in docker-gen 0.5.0 and the new version is 0.6.0.  I have just tested it and it is working fine. 

So now my nginx (official), docker-gen (0.6.0) and an LE companion container (modified with this pull request and my own modified nginx.tmpl) are working perfectly to grab/renew my SAN certificate under docker 1.10.1.

Thanks everyone! :) :+1:

nilroy:
@JrCs  Should I bump the docker-gen version here?

JrCs:
@nitroy not needed it's already in the master branch.

nilroy:
@JrCs  cool. I shall wait for this PR to be merged and the latest LE container available publicly. Till then I am using my locally built one.

JrCs:
@nitroy i'm currently thinking how to mix your PR and some enhancement i have.

nilroy:
@JrCs Well I think its better to do that in a separate PR

JrCs:
@falcon15500 What modifications have you done in nginx.tmpl.

falcon15500:
I am completely new to templating, but here is my nginx.tmpl - it's an amalgamation of the template inside your LE container and the one from the newer docker-gen (0.5.0 at the time), with it's support for the newer docker networking.


[nginx.tmpl.txt](https://github.com/JrCs/docker-letsencrypt-nginx-proxy-companion/files/132371/nginx.tmpl.txt)

JrCs:
@nilroy and @falcon15500 can you try my dev branch (image dev created also) ?

falcon15500:
@JrCs @nilroy I have tested out your DEV branch and it seems to be working fine.  I started fresh and it generated my 10+ domains worth under the 1 SAN certificate, sent a HUP to the nginx-gen which regenerated the nginx default.conf via my nginx.tmpl - all working.  Good job!

JrCs:
Thanks @falcon15500, need some little enhancement and il will merge into the master branch.

JrCs:
Close by 0d6d1052701dcb19faf208b1250d4f83b486b559

PR #28, opened 2016-02-25
Fix typo in environment variable

bradjones1:
Simple fix, but if you read things a little too literally like I did... you can get tripped up. :-)


JrCs:
Thanks !

PR #30, opened 2016-03-06
fix "Invalid character in DNS name"

Panthro:
the quotes are raising a "Invalid character in DNS name";

it took me some time to realise, so I hope noe more people will lose the same time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,8 +101,8 @@ For example
 
 ```bash
 $ docker run -d \
-    -e VIRTUAL_HOST="foo.bar.com,bar.com" \
-    -e LETSENCRYPT_HOST="foo.bar.com,bar.com" \
+    -e VIRTUAL_HOST=foo.bar.com,bar.com \
+    -e LETSENCRYPT_HOST=foo.bar.com,bar.com \
     -e LETSENCRYPT_EMAIL="foo@bar.com" ...
 ```
 ##### Automatic certificate renewal


JrCs:
Can you explain how to reproduce the problem ?
I have try the example in the readme and i don't see an "Invalid character in DNS name".

Panthro:
I ported the sample to a docker-compose file with the
```
ENVIRONMENT
 - VIRTUAL_HOST="foo.bar.com,bar.com" 
 - LETSENCRYPT_HOST="foo.bar.com,bar.com" 
```

And the error occurs, using docker run it won't happen because of the bash interpreter, but on a docker-compose it will.

I know it was entirely my fault, but as domains are now allowed to contain ` ` or any special chars I created the pull request so others won't have the same problem.

JrCs:
Thanks for the explaination.
But for the docker run command on a terminal it's better to have the strings quoted.

Panthro:
Well, I don't think anyone will use a domain with an invalid char, and the good thing is that bash will complain and warn the user (docker run will fail, easier to detect);

but up to you :ok_hand:

Panthro:
btw, sorry for the lack of explanation on the original PR description, It feld obvious to me, but clearly wasn't, my bad.

JrCs:
Close by 63b8458

Panthro:
even better! :+1: 
thanks

PR #56, opened 2016-05-01
Added $DH_PARAM_BITS evironment var

kije:
Gives the user of the image the ability to specify how strong the generated dhparams should be


briceburg:
LGTM -- any worry about merging?

JrCs:
Need to check if we can use "any number" for DH size.

JrCs:
@briceburg what is your use case to need to change the number of bits for DH keys. It's already 4096 and give use a A+ grade on  Qualsys SSL Server Test ?

JrCs:
No more info. Closing this PR

PR #58, opened 2016-05-03
Use `rw` in example in README

Dinoshauer:
Hey, just tried your examples in the README and of course I just copy/pasted things :)
`/etc/nginx/certs` needs to be mounted `:rw` so `jrcs/letsencrypt-nginx-proxy-companion` can save the certs :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ Example of use:
 ```bash
 $ docker run -d -p 80:80 -p 443:443 \
     --name nginx-proxy \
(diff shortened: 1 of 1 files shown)


JrCs:
Hi @Dinoshauer.
No the /etc/nginx/certs don't need to be mount rw in the nginx-proxy or nginx containers. This containers only read the certificates.
Only the companion need to have write access to the certs directory to write or update the certificates.

PR #61, opened 2016-05-23
enable public access to validation endpoints

benjamin-smith:
Enable public access to validation endpoints despite existing IP whitelisting or basic auth settings. For existing basic auth and IP whitelist setups, the `auth_basic off;` is not sufficient to allow Lets Encrypt to be able to access the container via HTTP.


JrCs:
@benjamin-smith thank for the PR.

Issue #63, opened 2016-05-26
How does well-known challenge work with default nginx.tmpl?


funkypenguin:
So I'm trying to migrate from https://github.com/dmp1ce/nginx-proxy-letsencrypt. I've built separate docker-gen and nginx containers. All works well, except the LE client can't validate my hostnames by serving the challenge-response from /usr/share/nginx/html

In the nginx template used in nginx-proxy-letsencrypt (https://github.com/dmp1ce/nginx-proxy-letsencrypt/blob/master/nginx.tmpl), I found the following:

	location /.well-known/ {
		root /usr/share/nginx/html;
		try_files $uri @proxy_pass;
	}

But there's no equivalent config in https://github.com/jwilder/nginx-proxy/blob/master/nginx.tmpl, as highlighted in your instructions.

As a result (I think), incoming challenges are served with a default http->https redirect to the proxied container, and the validation fails.

Is this expected, or have I missed a step somewhere? :)

Thanks!
D

funkypenguin:
Think I figured it out, submitted pull request:
https://github.com/JrCs/docker-letsencrypt-nginx-proxy-companion/pull/64

D

falcon15500:
This doesn't need to be (and shouldn't be) in the nginx.tmpl as we want it to stay stock.  The companion container provides this functionality via the vhosts.d/default entry. 

M.

JrCs:
Closed. See #64

funkypenguin:
Ah, thanks for the explanation. I'd managed (during the process of fiddling about) to start my 3 containers out-of-order, which meant that the nginx volume mounts didn't work, thus my problem.

PR #64, opened 2016-05-28
corrected instructions for nginx.tmpl, included local copy

funkypenguin:
nginx.tmpl requires additional config (from the docker-gen original) to support LetsEncrypt challenge-response. This request amends README.md to change instructions slightly, and includes amended nginx.tmpl as required.


JrCs:
Like @falcon15500 on #63 we must not change the nginx.tmpl file.
The challenge files are managed with vhosts.d/default: [nginx_location.conf](https://github.com/JrCs/docker-letsencrypt-nginx-proxy-companion/blob/master/app/nginx_location.conf)

PR #81, opened 2016-07-14
Removed "-only-exposed" so that "3-container setups" will work properly

joaocc:
With the "-only-exposed" option, when nginx-gen is a separate container, it no longer appears in the container list.
Due to the way the nginx.tmpl is setup, it will cause de detection of the CurrentContainer to fail.

Ideally, we should be able to fix the "nginx.tmpl" template, but this will solve most issues for now.


JrCs:
@joaocc but the _only-exposed_ option is used to ONLY detect change in "backend" web containers (that expose a/some port(s)).
So i don't understand why it can cause problem in the 3 containers configuration ?

joaocc:
Hi JrCs. 
I made the PR sometime ago and I don't really remember why I needed it, since my current setup uses your "unmodified" container.
So, I am going to close this PR to prevent confusion.
Cheers

PR #126, opened 2016-11-28
Support referencing nginx-gen container indirectly through another environment variable

ento:
### Motivation

I'm running this container alongside nginx-gen in Amazon ECS. The docker-gen container gets assigned a mangled name like "ecs-tws3-24-nginx-gen-aa9cf4babdd593ae5e00", which means I can't simply set `NGINX_DOCKER_GEN_CONTAINER=nginx-gen`.

What I ended up doing is [linking](https://docs.docker.com/engine/userguide/networking/default_network/dockerlinks/) those containers and having Docker assign and expose a namespaced name as an environment variable like:

```sh
NGINX_GEN_NAME=/ecs-tws3-23-nginx-letsencrypt-a6d9b293fc808bf7fd01/nginx-gen
```

### What this PR does

This PR adds the ability to specify the name of this environment variable instead of directly specifying docker-gen's container ID or name.

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,5 +150,7 @@ $ docker run -d \
 
 * `NGINX_PROXY_CONTAINER`- If for some reason you can't use the docker --volumes-from option, you can specify the name or id of the nginx-proxy container with this variable.
 
+* `NGINX_DOCKER_GEN_LINKED_CONTAINER_ENVVAR` - If you are running nginx-gen as a [linked container](https://docs.docker.com/engine/userguide/networking/default_network/dockerlinks/) and its ID cannot be predetermined, specify the name of the environment variable of the form `<alias>_NAME` that gets set by Docker when linking containers.
+
 #### Examples:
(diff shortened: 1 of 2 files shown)


JrCs:
@ento can you fix the code with the reviews i have made ?

JrCs:
I think the best is to use labels to references each others containers ?

teohhanhui:
I think this is closed by #181.

helderco:
@teohhanhui #181 is only for the joined `jwilder/nginx-proxy` container, not separate `jwilder/docker-gen` and `nginx`. 

But #231 is another approach using a label and should solve this as well.

ento:
Confirmed #231 works with Amazon ECS by using the `dockerLabels` property:

```
{
  {
    "image": "nginx",
    "dockerLabels": {
      "com.github.jrcs.letsencrypt_nginx_proxy_companion.nginx_proxy": ""
    },
    ...
  },
  {
    "image": "ento/docker-nginx-gen-with-tpl",
    "dockerLabels": {
      "com.github.jrcs.letsencrypt_nginx_proxy_companion.docker_gen": ""
    },
    ...
  },
  ...
}
```

PR #129, opened 2016-12-02
Added chain.pem plugin to simp_le call

DrMurx:
This allows other services which need the CA certificate in a dedicated
file (such as postfix or dovecot) to use the created certificates.


JrCs:
Closed by 8d3b188

Issue #130, opened 2016-12-08
DeserializationError: Deserialization error: Wrong directory fields


Dabada:
Got an error when starting a nodeBB virtual host container : DeserializationError: Deserialization error: Wrong directory fields

# STEPS
## Container nginx-proxy

```
docker run -d -p 80:80 -p 443:443 \
    --name nginx-proxy \
    -v /srv/nginx/certs:/etc/nginx/certs:ro \
    -v /etc/nginx/vhost.d \
    -v /usr/share/nginx/html \
    -v /var/run/docker.sock:/tmp/docker.sock:ro \
    jwilder/nginx-proxy
```

## Container nginx-letsencrypt

```
docker run -d \
    --name nginx-letsencrypt \
    -v /srv/nginx/certs:/etc/nginx/certs:rw \
    --volumes-from nginx-proxy \
    -v /var/run/docker.sock:/var/run/docker.sock:ro \
    jrcs/letsencrypt-nginx-proxy-companion
```

## Container node-bb

```
docker run -e "VIRTUAL_HOST=xxx.ovh.net" \
    -e "LETSENCRYPT_HOST=xxx.ovh.net" \
    -e "LETSENCRYPT_TEST=true" \
    --name node-bb \
    --link node-bb-redis:redis \
    -p 4567:4567 -P -t -i \
    benlubar/nodebb:latest
```

## Letsencrypt logs 

docker logs -f nginx-letsencrypt 

```
2016/12/08 13:01:38 Received event start for container 16ba23c35d3f
2016/12/08 13:01:53 Debounce minTimer fired
2016/12/08 13:01:53 Generated '/app/letsencrypt_service_data' from 3 containers
2016/12/08 13:01:53 Running '/app/update_certs'
Creating/renewal vps350014.ovh.net certificates... (vps350014.ovh.net)
2016-12-08 13:01:53,873:INFO:simp_le:1211: Generating new account key
2016-12-08 13:01:55,068:INFO:requests.packages.urllib3.connectionpool:756: Starting new HTTPS connection (1): acme-staging.api.letsencrypt.org
Traceback (most recent call last):
  File "build/bdist.linux-x86_64/egg/simp_le.py", line 1401, in main
    return main_with_exceptions(cli_args)
  File "build/bdist.linux-x86_64/egg/simp_le.py", line 1386, in main_with_exceptions
    persist_new_data(args, existing_data)
[...]

Issue #112, opened 2016-09-28
Disable basic auth for location /.well-known/acme-challenge/


ypereirareis:
Hi !
Actually, I have preprod environments protected by basic auth.
I would like to generate and renew certs without temporarily remove basic auth.

Maybe something like this:

```
location /.well-known/acme-challenge/ {
    auth_basic off
    allow all;
    root /usr/share/nginx/html;
    try_files $uri =404;
    break;
}
```

tmtben:
Hi @JrCs,
Thanks for your great work.
What about this feature?

buchdag:
Fixed by #232, closing the issue.

PR #132, opened 2016-12-09
Switch to simp_le's acme-0.8 branch before building

alastaircoote:
As discussed in #130, we're currently unable to issue SSL certs with this docker image. It's because simp_le needs to merge pull request [#112](https://github.com/kuba/simp_le/pull/112) and use a more recent version of the acme dependency. 

This is a short-term fix that manually switches to the branch of that pull request during build. This fix should be removed as soon as that pull request is merged, as it will be using a stale branch of the simp_le code. Unfortunately the last commit to simp_le was 8 months ago.


JrCs:
Closed by 8724387

Issue #134, opened 2016-12-12
Doesn't work anymore


cybergrunge:
It used to work like a charm, I have a “certs” folder filled with certs and keys for my apache containers, but when trying to add new containers it only generates the “domain.tld“ folder in the “certs” folder, without the certs and keys, and without the symlinks.

joeknock90:
I've got #128  open for it and someone opened #130 for a different breaking issue. I suggest posting logs just in case it's different than one of those.

danhimalplanet:
I think the current version of the python acme client breaks because as of Dec 8th 2016 the object returned by the /directory letsencrypt endpoint has changed

https://github.com/letsencrypt/boulder/issues/2415

cybergrunge:
The unexpected behaviour occurs at the first attempt to connect with https to a newly created webserver container (creation of the folder in “path/to/certs”, no generation of certs and keys). Where can I find logs of this precise action?

fayce:
its working again. I think letsencrypt made a modification in their API

smith64fx:
@fayce Really?

We still get this error
```
nginx-ssl    |   File "build/bdist.linux-x86_64/egg/simp_le.py", line 1401, in main
nginx-ssl    |     return main_with_exceptions(cli_args)
nginx-ssl    |   File "build/bdist.linux-x86_64/egg/simp_le.py", line 1386, in main_with_exceptions
nginx-ssl    |     persist_new_data(args, existing_data)
nginx-ssl    |   File "build/bdist.linux-x86_64/egg/simp_le.py", line 1282, in persist_new_data
nginx-ssl    |     client = registered_client(args, existing_data.account_key)
nginx-ssl    |   File "build/bdist.linux-x86_64/egg/simp_le.py", line 1224, in registered_client
nginx-ssl    |     client = acme_client.Client(directory=args.server, key=key, net=net)
nginx-ssl    |   File "build/bdist.linux-x86_64/egg/acme/client.py", line 63, in __init__
nginx-ssl    |     self.net.get(directory).json())
nginx-ssl    |   File "build/bdist.linux-x86_64/egg/acme/messages.py", line 169, in from_json
nginx-ssl    |     raise jose.DeserializationError(str(error))
nginx-ssl    | DeserializationError: Deserialization error: Wrong directory fields
```

jeisses:
[...]

PR #136, opened 2016-12-13
workaround #134 / #130

rossille:
## Disclaimer

I am perfectly aware that this is a dirty hack, and not good solution. But I think it's way better than the project not working at all.

## The problem

The root cause of the issue is already reported [on simpl_le](https://github.com/kuba/simp_le/issues/118)

It can be fixed by upgrading the client library. There is [a PR](https://github.com/kuba/simp_le/issues/118), but it's not merged yet.

## The workaround

Be cause I need this to work for my own setup, I forked docker-letsencrypt-nginx-proxy-companion to get the "acme-0.8" branch instead of "master" from simp_le

To prevent it from blowing up the day the branch is merged or removed, I added a fallback that gets the master if the branch doesn't exist.

This way, we can get docker-letsencrypt-nginx-proxy-companion back to work until the issue is fixed on simp_le and come back to remove this workaround at this time, and implement one of the better solutions discussed on the issue.

### DIFF
--- a/install_simp_le.sh
+++ b/install_simp_le.sh
@@ -7,7 +7,7 @@ apk --update add python py-requests py-setuptools git gcc py-pip musl-dev libffi
 
 # Get Let's Encrypt simp_le client source
 mkdir -p /src
-git -C /src clone https://github.com/kuba/simp_le.git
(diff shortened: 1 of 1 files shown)


JrCs:
Closed by daa88d29

PR #140, opened 2016-12-29
[FEATURE] Create SAN certificates only for common domains

mickaelperrin:
SAN certificates are now only created if the domain is contained in the base_domain.

For example:
     `LETSENCRYPT_HOST=domain.tld,sub.domain.tld,sub2.domain.tld,newdomain.tld,sub.newdomain.tld`

will create 2 SAN certificates for `domain.tld` and `newdomain.tld`.

### DIFF
--- a/app/letsencrypt_service
+++ b/app/letsencrypt_service
@@ -62,59 +62,83 @@ update_certs() {
         # First domain will be our base domain
         base_domain="${hosts_array_expanded[0]}"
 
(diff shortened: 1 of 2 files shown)


JrCs:
Thanks for this PR.

mickaelperrin:
Any problem with my PR ?

codev:
Why was this reverted? It seems like a really useful feature.

ladrua:
Yeah, this is definitely useful.

Issue #121, opened 2016-11-06
how do automated renewals work?


lvnilesh:
my cert is set to expire nov 12. 

I am expecting renewal to happen since Oct 12 but logs show
```
    Certificates already exist and renewal is not necessary, exiting with status code 1.
```
something seems wrong as I understand the renewal should have happened already but didn't.

What am I missing?

lvnilesh:
I am seeing this in the logs
```
2016-11-06 17:11:57,961:INFO:simp_le:1383: Certificates already exist and renewal is not necessary, exiting with status code 1.
Creating/renewal be.a.cloudgeni.us certificates... (be.a.cloudgeni.us)
Some plugins returned conflicting data for the "cert" component
```

lvnilesh:
What may be causing this conflicting data?

arcticShadow:
I've just noticed this as well. It appears that nginx has not restarted, and applied the new certs. As i checked the cert files and sure enough they have the correct expiration dates, indicating that this container did not need to renew the certs. (as the logs suggested)

prodrigestivill:
This is caused because docker-letsencrypt-nginx-proxy-companion only reloads docker-gen. And docker-gen don't see any change in the configuration files. So don't see any need to reload nginx.

I think it should also send reload to nginx container.

heinemml:
I stumbled accross the same problem.

**Oberservation**
 I also run the solution with an additional nginx-gen container. The letsencrypt-companion refreshes the certs and sends a notification to nginx-gen. And this is the output of nginx-gen:

````
2017/01/26 08:58:17 Received signal: hangup
2017/01/26 08:58:17 Received signal: hangup
2017/01/26 08:58:17 Contents of /etc/nginx/conf.d/default.conf did not change. Skipping notification ''
```` 

Which is correct, the configuration did not change. But the certs did, but this doesn't matter to docker-gen.

**Analysis**
So I guess the right way would be to also notify the nginx container when the certs changed.

I think this is even more important than notifying the docker-gen container. When a container is already proxied and has a cert generated the configuration shouldn't change at all. So the HUP will always lead to the situation mentioned above.

But the nginx container needs to be notified of every cert change.

**Workaround**
Workaround might be to set up a cron which HUPs your nginx container at least once a day.
```` docker kill -s HUP nginx````
 But a notification on update would of course be smoother.

buchdag:
This has been closed by #165

PR #165, opened 2017-02-08
Fix renewals on separate containers

prodrigestivill:
As described in issue #121 there is a bug when there is no new domains added, and needed to _SIGHUP_ _nginx_ container.
This pull requests automatize this reload in this case scenario.

It could be more optimal and only reload _nginx_ container in case no new domains had been found, but sending _SIGHUP_ to _nginx_ container twice doesn't seem like a big cost. At least for a quick short term solution.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ To run nginx proxy as a separate container you'll need:
 curl https://raw.githubusercontent.com/jwilder/nginx-proxy/master/nginx.tmpl > /path/to/nginx.tmpl
 ```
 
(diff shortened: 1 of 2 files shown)


heinemml:
Wouldn't it make sense to also use the 'NGINX_PROXY_CONTAINER' variable for this case? From a User-perspective I think it's a bit misleading to have two different NGINX container variables. The cases are mutual exclusive, so it should simplify the configuration.

prodrigestivill:
At first i tried to avoid the auto-fill to interfere with the NGINX variable. But it makes sense because the auto-fill only detects for a NGINX containter, and not for nginx-proxy.

buchdag:
@JrCs is there something preventing this from being merged ?

PR #168, opened 2017-02-13
Update README.md

ForsakenHarmony:
https://github.com/jwilder/nginx-proxy/pull/480/files#diff-5bd6b85c2d6fc987875f4bf82de2a15a

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ $ docker run -d \
     -v /path/to/nginx.tmpl:/etc/docker-gen/templates/nginx.tmpl:ro \
     -v /var/run/docker.sock:/tmp/docker.sock:ro \
     jwilder/docker-gen \
-    -notify-sighup nginx -watch -only-exposed -wait 5s:30s /etc/docker-gen/templates/nginx.tmpl /etc/nginx/conf.d/default.conf
+    -notify-sighup nginx -watch -wait 5s:30s /etc/docker-gen/templates/nginx.tmpl /etc/nginx/conf.d/default.conf
 ```
 
 * Then start this container (NGINX_DOCKER_GEN_CONTAINER variable must contain the docker-gen container name or id):


JrCs:
Hi, why do you want to remove the -only-exposed option ?
This option is used to select containers with exposed port(s) (which is the case if you want to server http requests).

oliverlorenz:
Same problem here. Same solution here. But it only works one time. 

If I start `docker-compose up` again the site does not appear. After I destroyed the containers with `docker-compose rm -f` and startup again with `docker-compose up -d` 

Any ideas?

thomasleveil:
TL;DR; in it's current state the jwilder/nginx-proxy template is not meant to be work with docker-gen and nginx being on separated container.

@JrCs since commit jwilder/nginx-proxy@658e20f661 the template in the jwilder/nginx-proxy projet make the assumption that the container running docker-gen is the container running nginx while [checking](https://github.com/jwilder/nginx-proxy/commit/658e20f661#diff-6c19a779564e1e2f88854a4a2fffbb78R83) that upstream containers are reachable. 

Since your docker-gen container does not expose any port, it's information won't be part of the data provided to the docker-gen process to fill up the template. As a result the `upstream` block is empty since the template is not able to check if the docker-gen container is on the same network as the upstream containers. (And that would make no sense anyway)

Removing the `-only-exposed` is a workaround. IMHO the best solution for your project is to provide and maintain your own template.

JrCs:
@thomasleveil thanks a lot for the study.
I really don't want to maintain my own version of the template.
Is it not possible to update the template do work in the 2 cases (one or two nginx-proxy/docker-gen containers) ?
If not, why the [nginx-proxy](https://github.com/jwilder/nginx-proxy) project do not support 2 versions of the template (it is said, in the readme, that the containers can be separated containers) ?

thomasleveil:
> Is it not possible to update the template do work in the 2 cases (one or two nginx-proxy/docker-gen containers) ?

Maybe, probably but that would require some thinking and work (maybe with a new environment variable `NGINXPROXY_CONTAINER`?)

> If not, why the nginx-proxy project do not support 2 versions of the template (it is said, in the readme, that the containers can be separated containers) ?

I don't know, that's a question for the maintainer of the project. But I doubt this would be feasible since it would mean more maintenance work and if you look at the number of opened issues and PR you can guess the maintainer is lacking the time. 

By the way, lot's of issues are coming from users of letsencrypt-nginx-proxy-companion, if you would just merge this PR, it would help preventing more flowing in.

thomasleveil:
jwilder/nginx-proxy#750 should provide the fix

JrCs:
Thanks @thomasleveil

Issue #170, opened 2017-02-13
Compatibility with docker-compose v3


joantune:
Hi!

First of all, kudos for the great work and container.

Currently, this container is not supported by docker-compose v3, because this new compose version has no `volumes_from` which is needed at:
https://github.com/JrCs/docker-letsencrypt-nginx-proxy-companion/blob/master/app/entrypoint.sh#L28

Any ideas on how to get the container's id any other way? I'm happy to do a PR if I have some pointers

I also asked that on an issue on docker-compose just now:

https://github.com/docker/compose/issues/4379#issuecomment-279499040

I don't know the rational for taking out a feature from compose, but I hope it's a good one on behalf of them, otherwise it kinda disencourages one to use it in the first place (IMO the docker engine and docker compose should have the same options possible)

For everyone else stuck on this, for now, just use v2.

joantune:
It seems that this the way to go:

https://github.com/docker/compose/issues/4379#issuecomment-279567234

I can try to use the volumes from, and if it fails, use the label. Should I proceed with a PR?

mikumi:
@joantune I am curious how you solved this, I haven't figured it out yet :)

JrCs:
@joantune yes you can proceed with a PR to use label instead of volumes_from to get the container id.

netdesk:
I did a setup using named volumes, without volumes_from. It works, don't know how as the entrypoint.sh of the letsencrypt container clearly uses VolumeFrom.

Here is the setup as a gist: https://gist.github.com/netdesk/c1db2985b542f9916995139318e5a7ce

The gist also includes a docker-compose setup for gitlab. One could include this in the nginx-proxy setup, but I wanted to separate concerns and the gitlab setup basically has nothing to do with the nginx-proxy thing.

PR #181, opened 2017-03-04
Get nginx container id from labelled container

emmetog:
Fixes https://github.com/JrCs/docker-letsencrypt-nginx-proxy-companion/issues/170

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ $ docker run -d -p 80:80 -p 443:443 \
     --label jrcs.nginx_letsencrypt_companion.is_proxy=true \
     jwilder/nginx-proxy
 ```
-The "jrcs.nginx_letsencrypt_companion.is_proxy" label is needed so that the letsencrypt container knows which nginx proxy container to use.
+The "jrcs.nginx_letsencrypt_companion.is_proxy=true" label is needed so that the letsencrypt container knows which nginx proxy container to use.
 
 * Second start this container:
 ```bash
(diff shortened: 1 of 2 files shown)


emmetog:
Not tested 100% yet, when I test it in docker compose v3 I'll post a comment. Feel free to comment on this PR in the meantime.

Tabrizian:
It's fully functional and completely compatible with docker compose v3. Thanks for your PR :+1:

README.md
```diff
@@ -30,14 +30,15 @@ $ docker run -d -p 80:80 -p 443:443 \
     -v /etc/nginx/vhost.d \
     -v /usr/share/nginx/html \
     -v /var/run/docker.sock:/tmp/docker.sock:ro \
+    --label jrcs.nginx_letsencrypt_companion.is_proxy=true \
```
teohhanhui: It should use a fully qualified domain name, i.e. `com.github.jrcs.letsencrypt_nginx_proxy_companion.nginx_proxy=true`

JrCs: Yes it will better to use a fully qualified domain name.
@emmetog can you make the changes ?

README.md
```diff
@@ -30,14 +30,15 @@ $ docker run -d -p 80:80 -p 443:443 \
     -v /etc/nginx/vhost.d \
     -v /usr/share/nginx/html \
     -v /var/run/docker.sock:/tmp/docker.sock:ro \
+    --label com.github.jrcs.letsencrypt_nginx_proxy_companion.nginx_proxy=true \
```
teohhanhui: There's no need for `=true`. Just a plain label is enough. It doesn't have to be a key-value pair.

PR #195, opened 2017-04-13
Use label for naming containers

### LINKED ISSUE #170: Compatibility with docker-compose v3

Tabrizian:
Tries to fix the problem of @emmetog pull request for using labeling instead of volumes.
Fixes #170

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ $ docker run -d -p 80:80 -p 443:443 \
     --label jrcs.nginx_letsencrypt_companion.is_proxy=true \
     jwilder/nginx-proxy
 ```
-The "jrcs.nginx_letsencrypt_companion.is_proxy" label is needed so that the letsencrypt container knows which nginx proxy container to use.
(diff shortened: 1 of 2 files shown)


JrCs:
Close by #181

PR #205, opened 2017-05-08
Added label-based mechanism to identify docker-gen container. This is needed w…

chicco785:
Added label-based mechanism to identify docker-gen container. This is needed when using docker compose or other container mechanisms that define dynamically container's name.

### DIFF
--- a/app/entrypoint.sh
+++ b/app/entrypoint.sh
@@ -45,6 +45,19 @@ function get_nginx_proxy_cid {
     fi
 }
 
+function get_nginx_gen_cid {
+    # Check if any container has been labelled as the nginx gen container.
+    local labeled_cid=$(docker_api "/containers/json" | jq -r '.[] | select( .Labels["com.github.jrcs.letsencrypt_nginx_proxy_companion.nginx_gen"] == "true")|.Id')
+    if [[ ! -z "${labeled_cid:-}" ]]; then
+        export NGINX_DOCKER_GEN_CONTAINER=$labeled_cid
+    fi
+    if [[ -z "${NGINX_DOCKER_GEN_CONTAINER:-}" ]]; then
+        echo "Error: can't get nginx-gen container id !" >&2
+        echo "Label the nginx gen container to use with 'com.github.jrcs.letsencrypt_nginx_proxy_companion.nginx_gen=true'." >&2
+        exit 1
+    fi
+}
+
 function check_writable_directory {
     local dir="$1"
     docker_api "/containers/$HOSTNAME/json" | jq ".Mounts[].Destination" | grep -q "^\"$dir\"$"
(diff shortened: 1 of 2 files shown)


app/entrypoint.sh
```diff
@@ -79,6 +92,7 @@ source /app/functions.sh
 
 if [[ "$*" == "/bin/bash /app/start.sh" ]]; then
     check_docker_socket
+    get_nginx_gen_cid
```
helderco: I think this should be: 
```
[[ -z "${NGINX_DOCKER_GEN_CONTAINER:-}" ]] && get_nginx_gen_cid
``` 
to allow still using a manual environment variable when you do have access to a static name.

app/entrypoint.sh
```diff
@@ -45,6 +45,19 @@ function get_nginx_proxy_cid {
     fi
 }
 
+function get_nginx_gen_cid {
+    # Check if any container has been labelled as the nginx gen container.
+    local labeled_cid=$(docker_api "/containers/json" | jq -r '.[] | select( .Labels["com.github.jrcs.letsencrypt_nginx_proxy_companion.nginx_gen"] == "true")|.Id')
+    if [[ ! -z "${labeled_cid:-}" ]]; then
+        export NGINX_DOCKER_GEN_CONTAINER=$labeled_cid
+    fi
+    if [[ -z "${NGINX_DOCKER_GEN_CONTAINER:-}" ]]; then
```
helderco: Remove this condition to make it optional, don't error out. Document it instead.

app/nginx_location.conf
```diff
@@ -1,4 +1,5 @@
-location ^~ /.well-known/acme-challenge/ {
+location ~ /.well-known {
```
helderco: Out of scope, changes should be discarded.

chicco785:
the change to `app/nginx_location.conf `may not be required. it's the results of different attempts to fix the issues i was facing in getting the certificate created. but the problem indeed was all related to the impossibility of connecting to the correct container due to the dynamic naming. the `com.github.jrcs.letsencrypt_nginx_proxy_companion.nginx_gen` label can be optional of course. let me know what you think.

helderco:
I need this, if you hadn't, I would! 👍

helderco:
I need this, if you hadn't, I would! 👍

helderco:
I found an issue with containers getting redeployed during letsencrypt lifecycle, so I made my own attempt with #231. It's working in production right now, with the [helder/letsencrypt-nginx-proxy-companion](https://hub.docker.com/r/helder/letsencrypt-nginx-proxy-companion/) fork (temporary until the PR is merged).

helderco:
#231 was merged so this can be closed.

JrCs:
Closed by #231

Issue #217, opened 2017-06-16
TLSA/HPKP: Any chance to use the existing key on renew?


narolinus:
Is there a way to use the existing key on certificate renewal? By default it seems to generate a new keypair on each renew. This renders TLSA- and HPKP-DNS-entries kind of useless. So there seems to be no easy way to use Public Key Pinning or TLSA together witch nginx-proxy-companion, instead one would need to update the entries manually, including a time window of failing certificate-checks for the length of the DNS-TTL.

buchdag:
That should be possible if `--reuse_key True` is passed to `simp_le`.

That's not possible right now without manually modifying [letsencrypt_service](../blob/master/app/letsencrypt_service#L112) but adding a new environment variable that would make it possible does not seems too hard.

narolinus:
That would be a great addition. I would really appreciate it.

buchdag:
PR opened.

buchdag:
Closed by #218

PR #218, opened 2017-06-16
enable simp_le private key reutilisation

buchdag:
This enable the use of simp_le `--reuse_key` option via the new environment variable `REUSE_KEY`

Reusing the previous private key at certificate renewal instead of generating a new one is almost mandatory if you want to setup things like [HPKP](https://developer.mozilla.org/en-US/docs/Web/HTTP/Public_Key_Pinning), as outlined in #217 

This commit has been tested on a real setup and is working as intended.

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,6 +159,8 @@ $ docker run -d \
 
 * `DEBUG` - Set it to `true` to enable debugging of the entrypoint script and generation of LetsEncrypt certificates, which could help you pin point any configuration issues.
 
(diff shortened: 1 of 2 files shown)


buchdag:
@JrCs  any issue with this PR ?

narolinus:
Tried it manually on my installation and can confirm that it does work. Great. Thank you.

buchdag:
You're welcome. I hope jrcs will have time to merge it soon.

buchdag:
@JrCs ?

buchdag:
Thank you !

PR #220, opened 2017-06-21
letsencrypt for other static sites

ymettier:
Hello,

docker-letsencrypt-nginx-proxy-companion is working well, with docker-gen, when you have web sites in a docker container in the same host as the nginx-proxy.

However, it could be extended to this use-case : I have another computer running some apache2 (it could be anything serving http on port 80). I want my nginx to reverse-proxy that web site too. And have nginx and docker-letsencrypt-nginx-proxy-companion doing the ssl. How could I do that ?

I analysed the container and I found it would be very easy to append some static data to /app/letsencrypt_service_data.

## Method 1

Write a file 

```
# Beginning of /app/letsencrypt_service_data.tmpl here
[...]
# then :
LETSENCRYPT_CONTAINERS+=( 'static_id1' )
LETSENCRYPT_static_id1_HOST=( 'my.example.com' )
LETSENCRYPT_static_id1_EMAIL="my.email@example.com"
LETSENCRYPT_static_id1_KEYSIZE="<no value>"
#LETSENCRYPT_static_id1_TEST=""
```

Then mount this file with docker -v /path/to/my/own/letsencrypt_service_data:/app/letsencrypt_service_data

With this method, there is nothing else to do. No Pull Request.

However, if I do that, I need to follow your /app/letsencrypt_service_data.tmpl and make mine any time you change yours.

I would prefer a mechanism to concatenate the _upstream_ file and my own file.

## Method 2

Have 2 files. The _upstream_ /app/letsencrypt_service_data.tmpl and my own file.

But it needs an change in /app/letsencrypt_service to do the concatenation. This is my Pull Request.

Thanks for helping proxify and ssl-ize web sites out of the docker world.

Regards,
Yves

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,6 +163,43 @@ $ docker run -d \
 
 * `ACME_TOS_HASH` - Let´s you pass an alternative TOS hash to simp_le, to support other CA´s ACME implentation.
 
+#### Static redirections
+
+You may want to define other servers than the ones in docker containers. You need to create a shell script file (that will be sourced) like this :
+
+```
+LETSENCRYPT_CONTAINERS+=( 'static_id1' )
+LETSENCRYPT_static_id1_HOST=( '<hostname>' )
+LETSENCRYPT_static_id1_EMAIL="<e-mail>"
+LETSENCRYPT_static_id1_KEYSIZE="<no-value>"
+#LETSENCRYPT_static_id1_TEST=""
+
+LETSENCRYPT_CONTAINERS+=( 'static_id2' )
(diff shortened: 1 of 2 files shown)


buchdag:
That won't work without modifications on `nginx-proxy` side as there is currently no feature to proxy to anything else than a Docker container IP short of [adding you own additional config file](https://github.com/buchdag/nginx-proxy#proxy-wide).

buchdag:
@ymettier your original idea was translated to something a bit more fleshed out and added to the `dev` branch for testing, any feedback is welcome on #368

PR #226, opened 2017-06-29
Update readme with sample using docker-compose version '3'

evertramos:
Hi there!

Thanks for the good work and I would like to share a new sample.

Cheers!

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,8 +165,9 @@ $ docker run -d \
 
 #### Examples:
 
-If you want other examples how to use this container, look at: 
+If you want other examples how to use this container, look at:
 
+* [Evert Ramos's Examples](https://github.com/evertramos/docker-compose-letsencrypt-nginx-proxy-companion) - using docker-compose version '3'
 * [Karl Fathi's Examples](https://github.com/fatk/docker-letsencrypt-nginx-proxy-companion-examples)
 * [More examples from Karl](https://github.com/pixelfordinner/pixelcloud-docker-apps/tree/master/nginx-proxy)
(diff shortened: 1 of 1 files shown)


buchdag:
@evertramos your docker-compose configuration won't reload nginx on automated certificate renewal and serve the renewed certificates due to the absence of the `NGINX_PROXY_CONTAINER` environment variable.

See the last bullet point [there](https://github.com/JrCs/docker-letsencrypt-nginx-proxy-companion#separate-containers-recommended-method) and [the relevant PR](https://github.com/JrCs/docker-letsencrypt-nginx-proxy-companion/pull/165).

evertramos:
Thanks for the comment @buchdag,

I have the option in the _jrcs/docker-letsencrypt-nginx-proxy-companion_ container, as of:

```
environment:
      NGINX_DOCKER_GEN_CONTAINER: "nginx-gen"
```

Isn´t that right? Or I must have specify it somewhere else?

buchdag:
If you use a docker-compose version 3 file with the three container setup (nginx / docker-gen / letsencrypt) you actually have to use both environment variable for the letsencrypt container:

`NGINX_PROXY_CONTAINER` pointing to the nginx container itself
`NGINX_DOCKER_GEN_CONTAINER` pointing to the docker-gen container as you did

evertramos:
@buchdag 

So it would be like this:

```
nginx-letsencrypt:
[...]
environment:
      NGINX_DOCKER_GEN_CONTAINER: "nginx-gen"
      NGINX_PROXY_CONTAINER: "nginx"
```

Is that it?

Thank you very much for your attention on that! It would take 90 days to realize it.

buchdag:
Yep, that's it !

PR #222, opened 2017-06-24
install simp_le from master branch + make that work, fixes #221

almereyda:
This has been written with some help from https://stackoverflow.com/a/19048754/1959568

I suggest to squash the commits, as the second description is misleading and should read something like *update dependencies and installation flow for simp_le, avoiding virtualenv*.

Issue #221, opened 2017-06-24
simp_le 0.2.0 is outdated


almereyda:
As configured by [`install_simp_le.sh`#L9-L11](https://github.com/JrCs/docker-letsencrypt-nginx-proxy-companion/blob/master/install_simp_le.sh#L9-L11), simp_le is installed from zenhack/simp_le:0.2.0, yet this branch does not exist anymore. [`master`](https://github.com/zenhack/simp_le/branches) seems to be a valid substitute.

almereyda:
Before I hadn't recognised the releases, which mention `0.2.0`.
Still the readme claims to be advanced to [`0.3.0`](https://github.com/zenhack/simp_le#030), for which no tag exists.

buchdag:
[simp_le 0.3.0 isn't released yet.](https://github.com/zenhack/simp_le/issues/57#)

PR #227, opened 2017-06-30
Build a specific release of simp_le instead of master

buchdag:
I think this container should aim at predictable build, at least for the core `simp_le`.

The removal of `--depth=1` is required to be able to make a `git checkout` and only represents a delta of 124 Kb as of today.

The removal of `--branch` makes git clone `master` by default.

This PR does not revert the improvements made by @almereyda on #222

### DIFF
--- a/install_simp_le.sh
+++ b/install_simp_le.sh
@@ -6,13 +6,13 @@ set -e
 apk --update add python py-setuptools git gcc py-pip musl-dev libffi-dev python-dev openssl-dev
 
 # Get Let's Encrypt simp_le client source
-branch="master"
 mkdir -p /src
-git -C /src clone --depth=1 --branch $branch https://github.com/zenhack/simp_le.git
+git -C /src clone https://github.com/zenhack/simp_le.git
 
 # Install simp_le in /usr/bin
+release='0.3.0'
 cd /src/simp_le
-#pip install wheel requests
+git checkout "$release"
 for pkg in pip distribute setuptools wheel
 do
   pip install -U "${pkg?}"


almereyda:
As much as I understand the reason of [*why* a shallow clone cannot target at a specific commit](https://stackoverflow.com/questions/26135216/why-isnt-there-a-git-clone-specific-commit-option), I would very much favour the social over the technical approach to rather ask the maintainer of `simp_le` to have shorter release cycles.

As you pointed out correctly in https://github.com/JrCs/docker-letsencrypt-nginx-proxy-companion/issues/221#issuecomment-311193619, proper tags didn't exist before. Luckily [`0.3.0` has been tagged 19 hours ago](https://github.com/zenhack/simp_le/releases) (from now).

Then, intermediate releases could follow semantic versioning and show up as `0.3.1`, `0.3.2` and the likes.

almereyda:
Also cloning the whole tree increases the image size of the container, if we don't run all `git` related commands within one `RUN` directive and the ` \` multiline pattern, including `rm -rf .git` after successful checkout.

buchdag:
You're more than welcome to go for the social approach and ask @zenhack to have shorter release cycles. [Actually I asked him to tag the 0.3.0 release 4 days ago](https://github.com/zenhack/simp_le/issues/57), but I don't feel like asking any more from him, not being a contributor to `simp_le`.

Meanwhile I'd really prefer this project to build against a fixed release of `simp_le`, but that's up to @JrCs, not me.

The `simp_le` installation is handled by a separate script, not by the Dockerfile. As for the sizes:

79285426 bytes for the latest image of this project built locally from a github clone.
79285333 bytes for the one I built with the commit in this PR (and after removing the first one) 
79285392 bytes for the last one I built with `git -C /src clone --depth=1 --branch '0.3.0' https://github.com/zenhack/simp_le.git` in the install script to avoid cloning the whole tree.

We're looking at size differences of less than 100 bytes, so I don't believe that cloning the whole tree in the `install_simp_le.sh` script actually increase the size of the container.  [The cloned source directory is being removed at the end of the script anyway.](https://github.com/JrCs/docker-letsencrypt-nginx-proxy-companion/blob/master/install_simp_le.sh#L24)

zenhack:
Fwiw:

* The release cycle is slow because development is slow -- I try not to let much time go by between a user-visible change and a release.
* per @buchdag's comment, the repo just isn't very big, and that's unlikely to change.
* simp_le is also available on pypi; any reason not to just pip install? This would obviate the questions about clone depth. Note I haven't really looked at your project, so maybe I'm missing something obvious here.

buchdag:
This project was originaly using kuba's simp_le, the switch to your fork was only made a month and a half ago. Maybe kuba's original project wasn't as well maintained on pypi as yours, hence jrcs choosing the git route instead.

zenhack:
I don't think the original hit pypi at all.

almereyda:
Yes, I overlooked the `install_simp_le.sh` script does all in one filesystem layer. Fine for me to clone the whole tree, especially with the evidence provided.
If we wanted to use a Docker build argument, we'd be free to choose the tag upon image creation.

JrCs:
Closed by #229

PR #231, opened 2017-07-08
Complete support for dynamic container names

helderco:
This is my attempt to solve the same #205 issue.

In Swarm Mode and Docker Cloud, container names are dynamic. #181 solved this partially with a label on `jwilder/nginx-proxy` but not if they're separate containers. #126 tries a different approach, but I think using a label is a better one.

This PR simply adds a `com.github.jrcs.letsencrypt_nginx_proxy_companion.docker_gen=true` label to the docker-gen container.

It also moves the label related detection to runtime (functions.sh: nginx-reload) instead of startup time (entrypoint.sh), because the nginx and docker-gen containers could be redeployed during the lifecycle of the letsencrypt companion, and thus, getting different ids.

There's also a new assumption in that I think it's simpler to just recommend using the label instead of `volumes-from` to detect the nginx-proxy container. The reason is that in some environments (like mine), you can't use `volumes-from`, but I think you can always use labels.

Note: If you're wandering why I've chosen not to set the variables `NGINX_DOCKER_GEN_CONTAINER` and `NGINX_PROXY_CONTAINER`, that's because we can use them to make sure an explicitly set container name takes precedence over a label.

Now I'm running a successful separate containers deployment in Swarm Mode, here's my stack:
```
version: '3.2'
services:
  nginx:
    image: nginx:alpine
    ports:
      - 80:80
      - 443:443
    volumes:
      - ./nginx/conf.d:/etc/nginx/conf.d
      - ./nginx/vhost.d:/etc/nginx/vhost.d
      - ./nginx/certs:/etc/nginx/certs
      - ./nginx/letsencrypt:/usr/share/nginx/html
    labels:
      - com.github.jrcs.letsencrypt_nginx_proxy_companion.nginx_proxy
    deploy:
      placement:
        constraints:
          - node.role == manager

  dockergen:
    image: helder/docker-gen
    command: -notify "docker-label-sighup com.docker.swarm.service.name=frontend_nginx" -watch /etc/docker-gen/templates/nginx.tmpl /etc/nginx/conf.d/default.conf
    volumes:
      - /var/run/docker.sock:/tmp/docker.sock:ro
      - ./nginx.tmpl:/etc/docker-gen/templates/nginx.tmpl
      - ./nginx/conf.d:/etc/nginx/conf.d
      - ./nginx/vhost.d:/etc/nginx/vhost.d
      - ./nginx/certs:/etc/nginx/certs
    labels:
      - com.github.jrcs.letsencrypt_nginx_proxy_companion.docker_gen
    deploy:
      placement:
        constraints:
          - node.role == manager

  letsencrypt:
    image: helder/letsencrypt-nginx-proxy-companion
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock:ro
      - ./nginx/conf.d:/etc/nginx/conf.d
      - ./nginx/vhost.d:/etc/nginx/vhost.d
      - ./nginx/certs:/etc/nginx/certs
      - ./nginx/letsencrypt:/usr/share/nginx/html
    deploy:
      placement:
        constraints:
          - node.role == manager
```

As you can see, I'm also using a custom `helder/docker-gen` image which simply adds a `docker-label-sighup` script that allows me to reload the nginx container from a label, because... well, dynamic names. `jwilder/docker-gen` doesn't support reloading containers with dynamic names (yet).

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ To run nginx proxy as a separate container you'll need:
 curl https://raw.githubusercontent.com/jwilder/nginx-proxy/master/nginx.tmpl > /path/to/nginx.tmpl
 ```
 
-2) Set the `NGINX_DOCKER_GEN_CONTAINER` environment variable to the name or id of the docker-gen container.
+2) Use the `com.github.jrcs.letsencrypt_nginx_proxy_companion.docker_gen=true` label on the docker-gen container, or explicitly set the `NGINX_DOCKER_GEN_CONTAINER` environment variable to the name or id of that container.
 
 Examples:
 
@@ -86,23 +86,25 @@ $ docker run -d \
     --volumes-from nginx \
     -v /path/to/nginx.tmpl:/etc/docker-gen/templates/nginx.tmpl:ro \
     -v /var/run/docker.sock:/tmp/docker.sock:ro \
+    --label com.github.jrcs.letsencrypt_nginx_proxy_companion.docker_gen=true \
     jwilder/docker-gen \
     -notify-sighup nginx -watch -wait 5s:30s /etc/docker-gen/templates/nginx.tmpl /etc/nginx/conf.d/default.conf
 ```
 
-* Then start this container (NGINX_DOCKER_GEN_CONTAINER variable must contain the docker-gen container name or id):
+* Then start this container:
 ```bash
 $ docker run -d \
     --name nginx-letsencrypt \
-    -e "NGINX_DOCKER_GEN_CONTAINER=nginx-gen" \
     --volumes-from nginx \
     -v /path/to/certs:/etc/nginx/certs:rw \
     -v /var/run/docker.sock:/var/run/docker.sock:ro \
     jrcs/letsencrypt-nginx-proxy-companion
 ```
-Then start any containers to be proxied as described previously.
 
-* If for some reason you can't use the docker --volumes-from option, you can specify the name or id of the nginx container with `NGINX_PROXY_CONTAINER` variable.
+* Then start any containers to be proxied as described previously.
+
+Note: If the docker-gen container name is static and you want to explicitly set it, use `-e NGINX_DOCKER_GEN_CONTAINER=nginx-gen`. The same thing is true with the nginx container (`-e NGINX_PROXY_CONTAINER=nginx`).
+
 
 #### Let's Encrypt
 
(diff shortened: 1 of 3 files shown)


buchdag:
Maybe the commits should be squashed to have a cleaner commit history and no doing-undoing commits ?

Otherwise nice PR, I hope it'll be merged.

helderco:
Agreed, but that's done [when merging](https://help.github.com/articles/about-pull-request-merges/#squash-and-merge-your-pull-request-commits). Afterwards, my fork will be deleted. 😺

README.md
```diff
@@ -63,7 +63,7 @@ To run nginx proxy as a separate container you'll need:
 curl https://raw.githubusercontent.com/jwilder/nginx-proxy/master/nginx.tmpl > /path/to/nginx.tmpl
 ```
 
-2) Set the `NGINX_DOCKER_GEN_CONTAINER` environment variable to the name or id of the docker-gen container.
+2) Use the `com.github.jrcs.letsencrypt_nginx_proxy_companion.docker_gen=true` label on the docker-gen container, or explicitly set the `NGINX_DOCKER_GEN_CONTAINER` environment variable to the name or id of that container.
```
teohhanhui: There's no need for `=true`. Just a plain label is enough. It doesn't have to be a key-value pair.

helderco: Actually there is, because it defaults to an empty string and I kept the logic from #181 where it checks for the value "true": `jq -r '.[] | select( .Labels["'$1'"] == "true")|.Id'`

We would have to change this query, I'm not opposed to it.

helderco: Actually I think it's better to disregard the value, since it doesn't matter: `jq -r '.[] | select( .Labels["'$1'"])|.Id'` 👍

teohhanhui: Actually, it's incredibly inefficient to fetch the list of *all* containers. The Docker Engine API allows filtering by labels.

helderco: I'll look into that!

teohhanhui: I think we can do that in another PR. For reference, this is my modified version: https://github.com/tripviss/tripviss-nginx-proxy/blob/0028116697aef28e9e7d5fec98fc791194e98e2b/docker/letsencrypt-companion/functions.sh#L31-L59

helderco: It's better for performance but also makes it easier to filter by simply a label or label=value. 

I improved my [docker-label-sighup](https://github.com/helderco/docker-gen/blob/master/docker-label-sighup), based on this. 

It's cooler now, because it allows me to do this: `-notify "docker-label-sighup com.docker.swarm.service.name=frontend_nginx"`

So I can reuse the labels that already exist.

sitepilot:
Works great, thanks! When will this be merged?

jasonchi38:
The communication is not working if you stack deploy a container on another node with the swarm. Has this problem been resolved?

PR #249, opened 2017-08-12
Add the ability to force the renewal of all certificates

buchdag:
This PR add a `force_renew` script in the /app folder.

The script force the renewal of all existing certificates via the `--valid_min` argument of `simp_le` (asking for a minimum validity of 90 days, which for LE certificates means they'll be renewed no matter what).


buchdag:
@JrCs any thought on this ?

Issue #266, opened 2017-10-12
Can't automatic renew certificates


ArtemisSaber:
got error logs like this
~~~
Creating/renewal www.semimajoraxis.com certificates... (www.semimajoraxis.com)
2017-10-11 18:24:29,789:INFO:simp_le:1213: Generating new account key
2017-10-11 18:26:39,077:ERROR:acme.challenges:311: Unable to reach http://www.semimajoraxis.com/.well-known/acme-challenge/CTehfImmbm55NlFJ5UnhBkE_EEr-j2_Ljh6QI405aCU: HTTPConnectionPool(host='www.semimajoraxis.com', port=80): Max retries exceeded with url: /.well-known/acme-challenge/CTehfImmbm55NlFJ5UnhBkE_EEr-j2_Ljh6QI405aCU (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7fddb4368190>: Failed to establish a new connection: [Errno 110] Operation timed out',))
2017-10-11 18:26:39,082:WARNING:simp_le:1304: www.semimajoraxis.com was not successfully self-verified. CA is likely to fail as well!
2017-10-11 18:26:39,221:INFO:simp_le:1314: Generating new certificate private key
2017-10-11 18:26:41,132:ERROR:simp_le:1272: CA marked some of the authorizations as invalid, which likely means it could not access http://example.com/.well-known/acme-challenge/X. Did you set correct path in -d example.com:path or --default_root? Is there a warning log entry about unsuccessful self-verification? Are all your domains accessible from the internet? Failing authorizations: https://acme-v01.api.letsencrypt.org/acme/authz/-WpaCqooC7HziH3jb_foG_xSy_oYmmeEce72cc7tKa4
Challenge validation has failed, see error log.
~~~

executing curl http://www.semimajoraxis.com/.well-known/acme-challenge/CTehfImmbm55NlFJ5UnhBkE_EEr-j2_Ljh6QI405aCU -v
works well as intended, output is shown below
~~~
curl http://www.semimajoraxis.com/.well-known/acme-challenge/CTehfImmbm55NlFJ5UnhBkE_EEr-j2_Ljh6QI405aCU -v
*   Trying 107.170.194.132...
* Connected to www.semimajoraxis.com (107.170.194.132) port 80 (#0)
[...]

PR #270, opened 2017-11-01
Fix forced renewal of certificates

buchdag:
Forced renewal with `/app/force_renew` currently renew only the first certificate it find on `letsencrypt_service_data` due to a `&& shift` that should not have been there in the first place (my mistake entirely).

@JrCs > could you please merge this as soon as convenient for you ? This feature is really needed for troubleshooting.

### DIFF
--- a/app/letsencrypt_service
+++ b/app/letsencrypt_service
@@ -70,7 +70,7 @@ update_certs() {
         params_d_str=""
         [[ $DEBUG == true ]] && params_d_str+=" -v"
         [[ $REUSE_KEY == true ]] && params_d_str+=" --reuse_key"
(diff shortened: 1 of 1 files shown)


Avd6977:
When will this be merged? Looks like it also might fix #266

buchdag:
Why would it fix a cert generation issue ? It's just a script forcing renewal of all existing certificates, nothing else. If you cert creation fails, `/app/force_renew` won't change it.

Avd6977:
"Can't automatic renew certificates" is the issue title. That's they issue we're having with our certs. Getting emails that they are about to expire when they should be auto-renewing. Sorry if there was confusion there.

buchdag:
@JrCs > thanks !

PR #276, opened 2017-11-15
add some documentation

revolunet:
Hi,

Thanks for this great project. Tried to understand how it works and produced a little schema that i can share with you. is it correct ?

Also added a note in the readme to warn users that the domain name must be functionnal for the setup to work properly.


revolunet:
schema source : https://docs.google.com/drawings/d/1tQhliw8ixWBD8ygtN1K6oTs8a7L-FFk2hfkLXUbqFqg/edit?usp=sharing

Issue #277, opened 2017-11-15
TOS hash mismatch


benne:
There is a new LE TOS, so the TOS hash is currently wrong, and failing.

https://letsencrypt.org/repository/

cpu:
Hi @benne,

I'm not a maintainer of this project but I believe the root of this problem is with the upstream `simp_le` ACME client that `docker-letsencrypt-nginx-proxy-companion` uses. I opened https://github.com/zenhack/simp_le/issues/85 to get this issue addressed, thanks for reporting!

buchdag:
While the issue is being fixed upstream you should use the `ACME_TOS_HASH` en var set to the new ToS hash : `cc88d8d9517f490191401e7b54e9ffd12a2b9082ec7a1d4cec6101f9f1647e7b`

benne:
Oh, great - thanks :) Fast answer!
For the time being, I put in the new TOS hash manually, and things seems to be working as expected.

talolard:
Wow can't believe this was had an open issue and workaround so quickly. 
Thanks!

buchdag:
[This issue has been fixed in `simp_le`](https://github.com/zenhack/simp_le/pull/88), a PR to remove the hardcoded ACME ToS hash from `docker-letsencrypt-nginx-proxy-companion` will be submitted as soon as a fixed tagged release of `simp_le` is available.

revolunet:
thanks for that great collaborative resolution !

buchdag:
PR is up.

Until it is merged, if you need a fixed container, do:

```
git clone --branch tos-fix https://github.com/buchdag/docker-letsencrypt-nginx-proxy-companion.git
cd docker-letsencrypt-nginx-proxy-companion
docker build -t jrcs/letsencrypt-nginx-proxy-companion:latest .
```

PR #279, opened 2017-11-15
Upgrade acme tos hash variable

matir91:
Fix issue https://github.com/JrCs/docker-letsencrypt-nginx-proxy-companion/issues/277


JrCs:
Fixed by #282

PR #281, opened 2017-11-16
added ONLY_EXPOSED env parameter

dj-hedgehog:
Rancher users need to be able to disable dockergens --only-exposed parameter to be able to have their containers detected. We have been using this approach for a year now issuing certificates for 25+ webservices.

The code is originally from @mcnilz (https://github.com/mcnilz/docker-letsencrypt-nginx-proxy-companion/commit/72b17994a175a6f0caf33e8fb4a8b674e18fe441)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,6 +4,7 @@ MAINTAINER Yves Blusseau <90z7oey02@sneakemail.com> (@blusseau)
 
 ENV DEBUG=false              \
 	DOCKER_GEN_VERSION=0.7.3 \
+	ONLY_EXPOSED=1 \
 	DOCKER_HOST=unix:///var/run/docker.sock
 
 RUN apk --update add bash curl ca-certificates procps jq tar && \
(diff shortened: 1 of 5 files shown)


buchdag:
@dj-hedgehog turns out we did not really need `--only-exposed` at all, so I believe #230 takes care of this ?

dj-hedgehog:
@buchdag I just deleted a certificate from one of my domains, started again with the latest `docker-letsencrypt-nginx-proxy-companion` and started `nginx-proxy` again but it did not renew the certificate.

I restarted again with my own version from this PR and it renewed the certificate in question.
So I still need it :)

Is there any other information I can provide?

buchdag:
Right, #230 removed the `-only-exposed` from `/app/start.sh` but not from the  `reload_nginx()` function on `/app/functions.sh`.

I think you can modify the PR to just remove the `-only-exposed`from line 101 on `/app/function.sh`. 

`-only-exposed` does not appear to be of any use anymore in the case of nginx-proxy + the le companion (as the two templates fed to docker-gen already filter containers through other means), so we probably won't need a feature to turn it back on.

dj-hedgehog:
I removed all usages of `ONLY_EXPOSED` I added and removed `-only-exposed` from `/app/function.sh` and `/app/start.sh`

dj-hedgehog:
Oops, I missed those. They are gone now.

Issue #288, opened 2017-11-24
certificates are removed each hour


richardbrinkman:
After the update of pull request #287 the docker-letsencrypt-nginx-proxy-companion container tries to remove all certs each hour by an `rm -rf /etc/nginx/certs`.

The corresponding log:
```
nginx_letsencrypt_1  | 2017-11-24T02:00:51.465837646Z rm: can't remove '/etc/nginx/certs': Resource busy
nginx_letsencrypt_1  | 2017-11-24T02:00:51.466698138Z rm: can't remove '/etc/nginx/certs': Resource busy
nginx_letsencrypt_1  | 2017-11-24T02:00:51.467441543Z rm: can't remove '/etc/nginx/certs': Resource busy
nginx_letsencrypt_1  | 2017-11-24T02:00:51.468107700Z rm: can't remove '/etc/nginx/certs': Resource busy
nginx_letsencrypt_1  | 2017-11-24T02:00:51.468902502Z rm: can't remove '/etc/nginx/certs': Resource busy
nginx_letsencrypt_1  | 2017-11-24T02:00:51.469662416Z rm: can't remove '/etc/nginx/certs': Resource busy
```

The only place where I see an `rm -rf` in the sourcecode is in app/letsencrypt_service. Apparently the `base_domain`, `hosts_array_expanded`, `hosts_array` and/or `host_varname` variable is set incorrectly.

buchdag:
I'm not witnessing this behaviour on any of my running container. I'm doing additional tests right now. It's true that we have a couple of dangerous `rm -rf` with unchecked variables that need to be secured, I'll work on that asap.

In the meantime please pull and use
`jrcs/letsencrypt-nginx-proxy-companion:v1.6`
or
`jrcs/letsencrypt-nginx-proxy-companion:stable`

buchdag:
I think I found why you are experiencing this. Do you happen to have a trailing comma on one of your `LETSENCRYPT_HOST` value, like` LETSENCRYPT_HOST=example.com,` ?

philipp65:
Hi,

I think I'm experiencing the same with the result
```
Creating/renewal xxx.yyy.zz certificates... (xxx.yyy.zz)
2017-11-24 11:34:14,210:INFO:simp_le:1538: Retrieving Let's Encrypt latest Terms of Service.
2017-11-24 11:34:16,281:INFO:simp_le:1356: Generating new account key
[...]

Issue #254, opened 2017-08-31
Trailing comma causes endless loop


brainyron:
If the LETSENCRYPT_HOST variable for a container has a trailing comma, the script does not detect this and attempts to add a certificate for an empty domain name.  This causes simp_le to fail, and the update_certs loop repeats endlessly.

buchdag:
I think that can be easily fixed with adjustment to letsencrypt_service_data.tmpl, I'll give it a try tomorrow.

buchdag:
My fix worked, but as I suspected the same trailing comma in the `VIRTUAL_HOST` variable cause `nginx-proxy` / `docker-gen` to generate an invalid conf file anyway:

`nginx-proxy    | 2017/09/01 14:06:46 [emerg] 1#1: invalid number of arguments in "upstream" directive in /etc/nginx/conf.d/default.conf:54`

So there is no point in trying to fix that if the nginx.tmpl isn't fixed first to strip the trailing comma from `VIRTUAL_HOST`.  nginx.tmpl is way more complex than letsencrypt_service_data.tmpl and I'm not good enough with the Go template package to come up with a fix yet.

buchdag:
So the issue is fixed for `LETSENCRYPT_HOST` by 4085f64, but since a trailing comma in `VIRTUAL_HOST` will still cause `nginx-proxy` / `docker-gen` to generate an invalid nginx config file, I think using trailing commas should still be avoided for now.

PR #289, opened 2017-11-24
Do not rm -rf with a trailing empty var

buchdag:
This fix safeguard against potential attempted forced deletions of `/etc/nginx/certs` when trying to do `rm -rf /etc/nginx/certs/$var` with a `var` that is empty or contains only spaces.

See #288

### DIFF
--- a/app/letsencrypt_service
+++ b/app/letsencrypt_service
@@ -83,7 +83,7 @@ update_certs() {
             acme_ca_uri="https://acme-staging.api.letsencrypt.org/directory"
             if [[ ! -f /etc/nginx/certs/.${base_domain}.test ]]; then
                 # Remove old certificates
-                rm -rf /etc/nginx/certs/${base_domain}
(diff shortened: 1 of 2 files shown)


buchdag:
Added real fix for #254 and #288 in addition to the `rm -rf` safeguard.

app/letsencrypt_service
```diff
@@ -83,7 +83,7 @@ update_certs() {
             acme_ca_uri="https://acme-staging.api.letsencrypt.org/directory"
             if [[ ! -f /etc/nginx/certs/.${base_domain}.test ]]; then
                 # Remove old certificates
-                rm -rf /etc/nginx/certs/${base_domain}
+                [[ ! -z "${base_domain// }" ]] && rm -rf /etc/nginx/certs/${base_domain}
```
JrCs: If prefer that you use [[ -n "${base_domain// }" ]] && ...

buchdag: Sure, that's cleaner. Done.

Issue #291, opened 2017-11-30
Allow one account key for all registrations.


almereyda:
Recently we ran into the rate limit for registrations per IP address.

- https://letsencrypt.org/docs/rate-limits/

Alternatively, it shall be possible to configure a single account key for all issuances.

- https://letsencrypt.org/docs/integration-guide/

How would such a setting require changes to the current companion?

buchdag:
I've already got this feature covered, I'll try to merge it in `dev` this week end.

buchdag:
A first version of the feature has been commited to `dev` : 4334d1593e841c8a16b013e565980351d1d25d57.

It's closely inspired by the way `certbot` handle account keys. [You can check the doc](https://github.com/JrCs/docker-letsencrypt-nginx-proxy-companion/tree/dev#acme-account-keys) and try it now by pulling and using `jrcs/letsencrypt-nginx-proxy-companion:dev`.

**Note:** if you use the `dev` version with an existing certificates volume/dir, on the next certificate(s) renewal it will replace corresponding account keys with a symlink to

`/etc/nginx/certs/accounts/acme-v01.api.letsencrypt.org/directory/default.json`

This new file will be copied from the account key of the first renewed certificate.

This will in turn cause a re-validation of all the other domains to the new common ACME account when their certificates are due for renewal. So be sure that your `nginx-proxy` stack is working properly and that you have no issues wit LE domain validation before using `jrcs/letsencrypt-nginx-proxy-companion:dev`, and backup you certificates volume.

You can also disable the feature entirely by setting the `REUSE_ACCOUNT_KEYS` to `false` on the letsencrypt container. This will make the container behave as it does currently in `master` (create a new ACME account for almost every new domain).

The automatic switch to a common ACME account key on renewal has been tested with success on two separate hosts running different services.

I pondered about having the whole feature be opt-in or opt-out, IMHO the current behaviour is generating enough issues on its own for this to be opt-out.

JrCs:
LGTM

buchdag:
Merged in master as of #298

Issue #263, opened 2017-10-04
Not a valid e-mail address


Algodrill:
Hi,
I use nginx-proxy on my docker with 2 websites.
I want set my websites on HTTPS protocol with letsEncrypt.

i set in my docker-compose on nginx proxy:
` letsencrypt-nginx-proxy-companion:
    restart: always
    image: jrcs/letsencrypt-nginx-proxy-companion
    container_name: letsencrypt-nginx-proxy-companion
    volumes_from:
      - nginx
    volumes:
      - "/var/run/docker.sock:/var/run/docker.sock:ro"
      - "./volumes/proxy/certs:/etc/nginx/certs:rw"
    environment:
      - NGINX_DOCKER_GEN_CONTAINER=nginx-gen`

and on my docker-compose website:
`sample-website:
    restart: always
    image: sample-website
    build: ./samples/website
    container_name: sample-website
    volumes:
      - "./volumes/nginx-sample-website/conf.d/:/etc/nginx/conf.d"
      - "./volumes/config/sample-website/config.js:/usr/share/nginx/html/config.js"
    environment:
      - VIRTUAL_HOST=samplewebsite.example.com
      - VIRTUAL_NETWORK=nginx-proxy
      - VIRTUAL_PORT=80
      - LETSENCRYPT_HOST=sample.example.com
      - LETSENCRYPT_EMAIL=email@example.com`

My address is a gmail domaine.

i look a logs and i get:
`2017/10/04 15:15:57 Contents of /app/letsencrypt_service_data did not change. Skipping notification '/app/update_certs'
2017/10/04 15:15:57 Received event die for container a6182eccc345
2017/10/04 15:15:58 Received event start for container a6182eccc345
2017-10-04 15:16:00,318:INFO:simp_le:1213: Generating new account key
ACME server returned an error: urn:acme:error:invalidEmail :: The provided email for a registration was invalid :: Error creating new registration :: not a valid e-mail address`

Do you have a solution for this ? Thank you !

cpu:
@Algodrill If you can share the external IP address you are running docker-compose from I can check the Let's Encrypt server-side logs to determine what email address was being rejected. That might make it easier to determine how to fix the problem.

cpu:
@Algodrill Thanks for sending your IP address. 

The server is receiving an invalid contact value: 
`mailto:\u003cno value\u003e`. 

[...]

PR #298, opened 2017-12-07
Merge dev into master

buchdag:
In addition to the improved handling of ACME account keys decribed in #291 , this PR:

- change the name of env var `REUSE_KEY` to `REUSE_PRIVATE_KEYS` to avoid confusion with the new `REUSE_ACCOUNT_KEYS`environment variable.
- make the the container behave the same when you set `LETSENCRYPT_TEST` to `true` on a proxyed container and when you set the `ACME_CA_URI `environment variable to the Let's Encrypt staging api endpoint. Previously certs generated when explicitly using the LE stating api endpoint where not really considered test certificates by the container.
- avoids `--email` being passed to `simp_le` with no address when `LETSENCRYPT_EMAIL`is not set, making the contact email truly optional (and preventing issues like #263)
- update the doc to take all that into account and add two notes about CAA DNS records and IPv6.

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,6 +172,17 @@ If needed, you can force a running letsencrypt-nginx-proxy-companion container t
 $ docker exec nginx-letsencrypt /app/force_renew
 ```
 
+##### ACME account keys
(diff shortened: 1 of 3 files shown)


JrCs:
lgtm

PR #321, opened 2018-01-26
Log error if nginx-proxy is not run

SmotrovaLilit:
I added log error if nginx-proxy is not run,  and i modified function docker_exec in case if docker can not create exec comand.

### DIFF
--- a/app/functions.sh
+++ b/app/functions.sh
@@ -8,6 +8,25 @@
 [[ -z "${END_HEADER:-}" ]] && \
  declare -r END_HEADER='## End of configuration add by letsencrypt container'
 
+function check_nginx_proxy_container_run {
+    local _nginx_proxy_container=$(nginx_proxy_container)
+    if [[ $(docker_api "/containers/${_nginx_proxy_container}/json" | jq -r '.State.Status') = "running" ]];then
+        return 0
+    fi
+
+    echo "$(date "+%Y/%m/%d %T"), Error: nginx-proxy container ${_nginx_proxy_container}  isn't running." >&2
+    return 1
+}
+
+function check_two_containers_case() {
+    local _docker_gen_container=$(docker_gen_container)
+    if [[ -n "${_docker_gen_container:-}" ]]; then  #case with 3 containers
+        return 1
+    fi
+
+    return 0
+}
+
 add_location_configuration() {
     local domain="${1:-}"
     [[ -z "$domain" || ! -f "${VHOST_DIR}/${domain}" ]] && domain=default
@@ -59,8 +78,11 @@ function docker_exec {
     local cmd="${2?missing command}"
     local data=$(printf '{ "AttachStdin": false, "AttachStdout": true, "AttachStderr": true, "Tty":false,"Cmd": %s }' "$cmd")
     exec_id=$(docker_api "/containers/$id/exec" "POST" "$data" | jq -r .Id)
-    if [[ -n "$exec_id" ]]; then
+    if [[ -n "$exec_id" && "$exec_id" != "null" ]]; then
         docker_api /exec/$exec_id/start "POST" '{"Detach": false, "Tty":false}'
+    else
+        echo "$(date "+%Y/%m/%d %T"), Error: can't exec command ${cmd} in container ${id}. Check if the container is running." >&2
+        return 1
     fi
 }
 
@@ -102,6 +124,7 @@ reload_nginx() {
             echo "Reloading nginx proxy (${_nginx_proxy_container})..."
             docker_exec "${_nginx_proxy_container}" \
                         '[ "sh", "-c", "/usr/local/bin/docker-gen /app/nginx.tmpl /etc/nginx/conf.d/default.conf; /usr/sbin/nginx -s reload" ]'
+            [[ $? -eq 1 ]] && echo "$(date "+%Y/%m/%d %T"), Error: can't reload nginx-proxy." >&2
         fi
     fi
 }
--- a/app/letsencrypt_service
+++ b/app/letsencrypt_service
@@ -43,6 +43,9 @@ create_links() {
 }
 
 update_certs() {
+
+    check_two_containers_case && (check_nginx_proxy_container_run || return)
+
     [[ ! -f "$DIR"/letsencrypt_service_data ]] && return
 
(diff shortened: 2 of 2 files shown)


app/functions.sh
```diff
@@ -8,6 +8,25 @@
 [[ -z "${END_HEADER:-}" ]] && \
  declare -r END_HEADER='## End of configuration add by letsencrypt container'
 
+function check_nginx_proxy_container_run {
+    local _nginx_proxy_container=$(nginx_proxy_container)
+    if [[ $(docker_api "/containers/${_nginx_proxy_container}/json" | jq -r '.State.Status') = "running" ]];then
+        return 0
+    fi
+
+    echo "$(date), Error: nginx-proxy container ${_nginx_proxy_container}  isn't run" >&2
```
buchdag: `date` output should be formatted like docker-gen (`date "+%Y/%m/%d %T"`).

Minor english typo too: "nginx-proxy container `id` isn't running."

app/functions.sh
```diff
@@ -59,8 +78,11 @@ function docker_exec {
     local cmd="${2?missing command}"
     local data=$(printf '{ "AttachStdin": false, "AttachStdout": true, "AttachStderr": true, "Tty":false,"Cmd": %s }' "$cmd")
     exec_id=$(docker_api "/containers/$id/exec" "POST" "$data" | jq -r .Id)
-    if [[ -n "$exec_id" ]]; then
+    if [[ -n $exec_id && ($exec_id != "null") ]]; then
         docker_api /exec/$exec_id/start "POST" '{"Detach": false, "Tty":false}'
+    else
+        echo "$(date), Error: can't exec command ${cmd} in container ${id}. Check container is run" >&2
```
buchdag: `date` output formatting + typo: "Check if the container is running."

app/functions.sh
```diff
@@ -102,6 +124,7 @@ reload_nginx() {
             echo "Reloading nginx proxy (${_nginx_proxy_container})..."
             docker_exec "${_nginx_proxy_container}" \
                         '[ "sh", "-c", "/usr/local/bin/docker-gen /app/nginx.tmpl /etc/nginx/conf.d/default.conf; /usr/sbin/nginx -s reload" ]'
+            [[ $? -eq 1 ]] && echo "$(date), Error: can't reload nginx-proxy. Check container is run" >&2
```
buchdag: Same as above.

buchdag:
Hi ! Thanks for the PR. Were you able to find exactly why the `check_nginx_proxy_container_run` is failing with the three containers setup ?

app/functions.sh
SmotrovaLilit: Fixed

app/functions.sh
SmotrovaLilit: Fixed

app/functions.sh
SmotrovaLilit: Fixed

SmotrovaLilit:
Hi @buchdag

In case with three containers nginx_proxy_container is empty because function get_nginx_proxy_cid dosn't run: https://github.com/JrCs/docker-letsencrypt-nginx-proxy-companion/blob/master/app/entrypoint.sh#L94

buchdag:
... and the test are currently running with `--volumes-from`, so indeed you don't get an nginx container id on three container setup. Make sense.

I'll cleanup the "get nginx container id" logic later to make this work with the three containers setup too, that was planned anyway.

I've got only two minor request for changes that I missed the first time.

app/functions.sh
```diff
@@ -59,8 +78,11 @@ function docker_exec {
     local cmd="${2?missing command}"
     local data=$(printf '{ "AttachStdin": false, "AttachStdout": true, "AttachStderr": true, "Tty":false,"Cmd": %s }' "$cmd")
     exec_id=$(docker_api "/containers/$id/exec" "POST" "$data" | jq -r .Id)
-    if [[ -n "$exec_id" ]]; then
+    if [[ -n $exec_id && ($exec_id != "null") ]]; then
```
buchdag: The parenthesis here are not needed.

I'd prefer `[[ -n "$exec_id" && "$exec_id" != "null" ]]` to stay in line with existing code style.

app/functions.sh
```diff
@@ -102,6 +124,7 @@ reload_nginx() {
             echo "Reloading nginx proxy (${_nginx_proxy_container})..."
             docker_exec "${_nginx_proxy_container}" \
                         '[ "sh", "-c", "/usr/local/bin/docker-gen /app/nginx.tmpl /etc/nginx/conf.d/default.conf; /usr/sbin/nginx -s reload" ]'
+            [[ $? -eq 1 ]] && echo "$(date "+%Y/%m/%d %T"), Error: can't reload nginx-proxy. Check if the container is running" >&2
```
buchdag: I think the "Check if the container is running" on this error message is redundant, as we'll get it from the `docker_exec` error just before that, right ?

app/functions.sh
```diff
@@ -8,6 +8,25 @@
 [[ -z "${END_HEADER:-}" ]] && \
  declare -r END_HEADER='## End of configuration add by letsencrypt container'
 
+function check_nginx_proxy_container_run {
+    local _nginx_proxy_container=$(nginx_proxy_container)
+    if [[ $(docker_api "/containers/${_nginx_proxy_container}/json" | jq -r '.State.Status') = "running" ]];then
+        return 0
+    fi
+
+    echo "$(date "+%Y/%m/%d %T"), Error: nginx-proxy container ${_nginx_proxy_container}  isn't running" >&2
```
buchdag: Could you end the error messages with a `.` ? 😃

SmotrovaLilit: Of course)

app/functions.sh
SmotrovaLilit: Fixed.

app/functions.sh
SmotrovaLilit: Yes, you are right.

buchdag:
Thanks for the changes.

Looks good to me, can you squash the three commits into one or do you want me to do it ?

SmotrovaLilit:
Thanks for reviewing. I’ve squashed commits into one

PR #331, opened 2018-02-15
Added new example usage reference with docker-compose

progress44:


buchdag:
Correct if I'm wrong, but due to [this](https://github.com/blimpair/loadbalancer/blob/181212c6eb59f9b3836e2e87f75923538bf005e1/docker-compose.yml#L60) and [this](https://github.com/blimpair/loadbalancer/blob/181212c6eb59f9b3836e2e87f75923538bf005e1/.env#L3), your compose file example does not actually use `jrcs/letsencrypt-nginx-proxy-companion` but your own version ([`registry.progress44.com/letsencrypt`](https://github.com/blimpair/docker-letsencrypt-nginx-proxy-companion/tree/f534e429438925c457b76bc9be3c55d32e505651)), with no  ?

Same thing for `nginx` and `docker-gen` ?

buchdag:
@jrcs as stated this compose file does not use `jrcs/letsencrypt-nginx-proxy-companion` but a (now out of sync) fork. I think this is unclear and confusing, and shouldn't be merged. Any advice ?

JrCs:
@buchdag so we can close it ?

buchdag:
Yup.

PR #334, opened 2018-02-20
Merge dev into master (v1.8)

buchdag:
Changes in this PR:

- Code cleanup / style consistency / ambiguous or similar names 3af5ae8 and 3b212f2

- Update docker-gen to 0.7.4 9570237

- Update simp_le to 0.7.0 400a7da

this version of simp_le removes everything related to the ToS hash and just agree to the ToS by default, so the everything related to the ACME_TOS_HASH environment variable has been removed and a deprecation message is printed to log when it's used.

- Rework of the logic behind nginx-proxy and docker-gen container ID acquisition as it was not making a lot of sense (in addition to having bits here and there). 866dda1 16f0222

**nginx container**: try to get the id from label first, then the env var, then `--volumes_from` as last resort.

**docker-gen container**: try to get the id from label first, then the env var.

**at container startup**: check if a nginx container ID can be found, err exit if not, then check if a docker-gen container ID can be found **OR** if the nginx container appears to be running docker-gen (meaning it's actually an nginx-proxy container), err exit if not.

- Remove symlinks in /etc/nginx/certs when the corresponding domain is no longer used in any running container `LETSENCRYPT_HOST` environment variable. 4b2b442

This will be completed with notes on the v1.8 release and on the wiki about HSTS being enabled by default on nginx-proxy, and how you have to disable it first if you want to regain the ability to "switch" SSL on and off for a given proxyed service.


buchdag:
Additional commits have been pushed to the `dev` branch:

- The single domain certs whose domain is now part of another SAN certificate are no longer removed. There is no need to remove them as they are not symlinked to anyway, and keeping them allows going back to the single cert if required without issuing a new one. 23e214b

- The symlinks are no longer created / changed when `simp_le` was unable to generate the certificate. With the previous commit, this prevent issues like #337 where the target certificate does not cover the required domain(s) but the symlink get created anyway. fc0d685

- Rather than overwriting the certificates coming from the staging endpoint with those coming from the production endpoint and (more annoyingly) vice versa, both are now kept, the staging cert / key / chain being put in their own directory prefixed with `_test_`. 10f6f31

- Addition of a script that print information about existing certificates, including the domains that are using it. c823abe

PR #335, opened 2018-02-20
Add ^~ to location directive

thomaco:
This is a small change to make the matching a little bit faster, while also allowing more general blocking.

First some reference to the docs.
> If the longest matching prefix location has the “^~” modifier then regular expressions are not checked. 

[Nginx http_core_module](https://nginx.org/en/docs/http/ngx_http_core_module.html#location)
Further, from the docs, the evaluation order is the following;
1. Exact matches (using the = modifier)
2. Implicit prefix matching(no modifier or ^~)
3. Regex(~ and ~*)
 
As such, without a modifier(such as ^~), the location directive matches with the prefix of the challenge address, but can be overridden by a subsequent regex directive. An example is a general regex block for dot-files and dot-folders, such as this example taken from nextcloud;
> location ~ ^/(?:\.|autotest|occ|issue|indie|db_|console) {

By adding the ^~ modifier, we tell nginx to stop at step 2 and use that location block, and not look any further.

From what I've seen, this was removed earlier with the reason 
>Don't need to use a regexp because the vhost.d/default configuration must be include specificaly in each server configurations by the nginx.tmpl template file.

This is fine for hosts which are proxied straight through the nginx proxy, however, for hosts with custom configuration(such as a fpm container mounted on the nginx proxy), vhost.d/default is not included, and the directive is instead inserted at the top of the configuration file for that vhost definition which probably contain other location directives. I would also like to emphasize that ^~ is not a regex-modifer, it tells nginx to _not_ look at regex locations.

I hope it's okay to just send a pull request like this, and looking forward to any input.

### DIFF
--- a/app/nginx_location.conf
+++ b/app/nginx_location.conf
@@ -1,4 +1,4 @@
-location /.well-known/acme-challenge/ {
+location ^~ /.well-known/acme-challenge/ {
     auth_basic off;
     allow all;
     root /usr/share/nginx/html;


buchdag:
Hi

I think that the reason why it was removed was rather that @jrcs encountered errors with it, check the commit comments : https://github.com/JrCs/docker-letsencrypt-nginx-proxy-companion/commit/1f678ed2c205b94d0f66dc8ec0270805d86f74a4

I personally never encountered any error with it and it seemed to solve an issue at some point, but the same issue was gone even after the revert, so I don't really know where we stand with this right now, and I'm not experienced enough with nginx to see why the `^~` could have caused errors.

thomaco:
Hello,

Yes, I saw the commit, should have referenced it. The commit references an if statement in the nginx.tmpl, which means one of two cases. Here is how I see the two cases being affected by this commit;
(For both cases, the function add_location_configuration() in app/functions.sh will prepend the contents of nginx_location.conf to the vhost file)

1. A vhost.d/<host> file exists for the host
The missing ^~ modifier can cause problems, as a secondary regex match(like the example above) will override the location directive inserted by the letsencrypt container. The other problem situation I see is where another prefix location directive exists which has the same prefix as nginx_location.conf(here, I think nginx will complain), or a longer prefix match - both of which implies that someone else is handling acme-challenges for the container, so this container should probably not handle certificates for that container.

2. The vhost.d/default file exists
For the case where there is no default vhost configuration specified, this should not cause any problems, as only requests to /.well-known/acme-challenge/ will be routed differently - the rest hit the 'location /' directive. When there exists a default vhost configuration, the same applies as for alternative 1).

I am experiencing a problem with this location directive, and it can be seen in my [docker-nginx-nextcloud-grav](https://github.com/thomaco/docker-nginx-nextcloud-grav) repository. The fpmcron/nextcloud.example.com file is an example of a vhost.d/<host> file. It contains a location directive meant to deny access to dot-files and dot-folders, and since it's a regex match, it overrides the /.well-known/acme-challenge/ directive inserted into the top of the file by the letsencrypt companion.

streamholder:
Hello,

I'm detailing here, as requested by @buchdag (by the way, thank you for the help! :), my findings about the vhost.d/default problem.

To quickly reproduce, download and extract the archive, change the [hosts] string with actual hostnames in `docker-compose.yml` and run:
```
docker network create nginx-proxy-net
docker-compose up
```
[docker.tar.gz](https://github.com/JrCs/docker-letsencrypt-nginx-proxy-companion/files/1746177/docker.tar.gz)
(I tried to make it as simple as possible, it should be about the bare minimum needed to reproduce)

When going up, the cert generation will result in the .well-known requests being relayed upstream by the proxy, instead of being served by it. This seems to be because when docker-gen is generating the configuration, the file `/etc/nginx/vhost.d/default` doesn't yet exist, resulting in the location blocks for /.well-known/acme-challenge/ not being generated at all.

The location file I provided in the archive shows this by adding the following debug comment to the server blocks in the default.conf proxy configuration file generated by the docker-gen container:
`    # here shall lie the companion's evil device < {{ (printf "/etc/nginx/vhost.d/%s" $host) }} > < {{ exists "/etc/nginx/vhost.d/default" }} >`

Printing the generated configuration file shows that during the generation the /etc/nginx/vhost.d/default file indeed doesn't exist:
```
root@hostname:/srv/docker-website-bug# docker exec -ti proxy bash
root@b49181cc1556:/# cat /etc/nginx/conf.d/default.conf

[omissis]

server {
    server_name _; # This is just an invalid value which will never trigger on a real hostname.
    listen 80;
    access_log /var/log/nginx/access.log vhost;
    return 503;
}
upstream [host] {
   # Container: upstream@nginx-proxy-net
   server 172.20.0.2:80;
}
server {
    server_name [host];
    listen 80 ;
    access_log /var/log/nginx/access.log vhost;
    # here shall lie the companion's evil device < /etc/nginx/vhost.d/[host] > < false > [<-- here!]
    location / {
        proxy_pass [host];
    }
}
root@b49181cc1556:/#
```

What instead would be expected is:
```
[omissis]

server {
    server_name _; # This is just an invalid value which will never trigger on a real hostname.
    listen 80;
    access_log /var/log/nginx/access.log vhost;
    return 503;
}
upstream [host] {
   # Container: upstream@nginx-proxy-net
   server 172.20.0.2:80;
}
server {
    server_name [host];
    listen 80 ;
    access_log /var/log/nginx/access.log vhost;
    # here shall lie the companion's evil device < /etc/nginx/vhost.d/[host] > < true >
    include /etc/nginx/vhost.d/default;

    location / {
        proxy_pass [host];
    }
}
```

It is worth noting that at some point the `/etc/nginx/vhost.d/default` file gets created with the right content:
```
root@b49181cc1556:/# cat /etc/nginx/vhost.d/default
## Start of configuration add by letsencrypt container
location /.well-known/acme-challenge/ {
    auth_basic off;
    allow all;
    root /usr/share/nginx/html;
    try_files $uri =404;
    break;
}
## End of configuration add by letsencrypt container
root@b49181cc1556:/#
```

Indeed, I verified that hardcoding the expected snippet in the template file (i.e. removing the conditional surrounding it or just copy-pasting the contents of `/etc/nginx/vhost.d/default` in its place) results in the signing process working correctly. (Thank you @buchdag for the suggestion)

_I'm probably forgetting to write something, as it's late and I have a plane to catch tomorrow. You can try to contact me by email if you need clarifications and other information._

(In a mostly unrelated note, the nginx.tmpl linked in the README.md of this repo does _not_ work natively with this kind of setup, instead it results in empty upstream blocks. It's something you or someone else might want to look into.)

thomaco:
Hello,

Interesting problem, but to be clear; this doesn't seem to be related to this pull request.

Running `docker-compose up -d && docker-compose logs -f`, I can see the following when starting your setup;
```
companion    | /etc/nginx/certs/example.com /app
companion    | Reloading nginx docker-gen (using separate container proxy-gen)...
companion    | 2018/02/22 13:41:49 Contents of /app/letsencrypt_service_data did not change. Skipping notification '/app/update_certs'
companion    | Creating/renewal example.com certificates... (example.com www.example.com)
```

This should look like this;
```
companion              | Reloading nginx docker-gen (using separate container proxy-gen)...
proxy-gen                | 2018/02/22 13:44:15 Received signal: hangup
proxy-gen                | 2018/02/22 13:44:15 Received signal: hangup
companion              | Reloading nginx (using separate container nginx)...
companion              | Creating/renewal example.com certificates... (example.com www.example.com)
```

Alas, the docker-gen process never receives the SIGHUP signal!

The cause of this seems to be your cmd/entrypoint tuple, which causes two processes inside the container; one `bash -c` process(the root process), and one `/usr/local/bin/docker-gen`. The solution is simply to replace the value of entrypoint with the value of command, like this;
```
entrypoint: /usr/local/bin/docker-gen -notify-sighup proxy -watch -only-exposed -wait 5s:30s /etc/docker-gen/templates/nginx.tmpl /etc/nginx/conf.d/default.conf
```

Now, as to the why, I don't know enough about bash internals to state anything definitive. However, my theory is that the SIGHUP signal sent to the docker-gen container isn't propagated through the bash process to the actual docker-gen process, which leads to no new/updated nginx configurations getting generated.

thomaco:
Also, to add, nginx.tmpl from the Readme seems to work fine when entrypoint is correctly defined.

streamholder:
Hello,

I defined the entry point that way for the reason described here: https://github.com/jwilder/docker-gen/issues/201
Defining the entry point of docker-gen as suggested breaks my setup as well, and it does so consistently (that's the reason I never had a chance to notice the original nginx.tmpl works fine).

I posted under this particular GitHub issue as requested by Nicolas. I'm sorry if it was not related.

R.


Il 22 feb 2018 17:02, Thomas Oddsund <notifications@github.com> ha scritto:

Also, to add, nginx.tmpl from the Readme seems to work fine when entrypoint is correctly defined.

buchdag:
@thomaco my bad, I thought this might have been related and a broader issue with the `/.well-known/acme-challenge/`.

thomaco:
@streamholder for what setup does changing the entrypoint break your setup? Asking since the setup in the tarball you commented with started working after I swapped out the domains to my domain + changed the entrypoint. Would be interesting to try to debug.
Also, what exactly do you mean with "breaks my setup"? Container exits with error code 2 and no message, or something else?

@buchdag No worries buchdag. While this issue might indeed be related to some broader issue that I'm currently not aware of, the change suggested should only stop regex-directives from matching '/.well-known/acme-challenge/'.

streamholder:
Container exits with code 2 and no message. I haven't tried that with the specific setup I attached, but it did not work with the original setup I derived that one from.

(I'd really like to provide more details but as I explained I'm not home.)

R.

Il 22 feb 2018 18:05, Thomas Oddsund <notifications@github.com> ha scritto:

@streamholder<https://github.com/streamholder> for what setup does changing the entrypoint break your setup? Asking since the setup in the tarball you commented with started working after I swapped out the domains to my domain + changed the entrypoint. Would be interesting to try to debug.
Also, what exactly do you mean with "breaks my setup"? Container exits with error code 2 and no message, or something else?

@buchdag<https://github.com/buchdag> No worries buchdag. While this issue might indeed be related to some broader issue that I'm currently not aware of, the change suggested should only stop regex-directives from matching '/.well-known/acme-challenge/'.

buchdag:
@thomaco I'm not forgetting this PR, I'll get back to it as soon as #334 is merged and v1.8 released.

buchdag:
So. Using the `^~` make sense to.

Do you see a case where having it would have resulted in a failed authorization and removing it would have solved the issue ? I'll try to have @JrCs chime in on this but I believe he did not find exactly why the `^~` caused authorization issues in his case (authorization issues that appeared to be solved by reverting the commit).

thomaco:
Hello,

Not really. If an authorization fails because of ^~, that would imply that the authorization is dependent on a regex(since step 3 is never checked). That would mean that there is another location block handling acme challenges, which I guess doesn't make sense if one is using this service.

buchdag:
Ok, as far as I'm concerned I think this should be merged back, but it won't be without @JrCs approval.

PR #378, opened 2018-05-12
Enable tests on macOS

buchdag:
This PR enable local testing on macOS by taking into account minor differences with `sed`, `getopt` and `bash` versions packaged with this OS.

I had to reverted `10.77.77.1` to `localhost` in `test/setup/setup-boulder.sh`, as using the Docker gateway IP to curl Boulder's ACME endpoint from macOS Terminal does not appear to work.

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -1,6 +1,6 @@
 ### letsencrypt-nginx-proxy-companion test suite
 
-The test suite can be run locally on a Linux host.
+The test suite can be run locally on a Linux or macOS host.
 
 To prepare the test setup:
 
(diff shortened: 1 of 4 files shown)


buchdag:
I modified the ACME endpoint check again to have it [performed from within the Boulder container itself](https://github.com/JrCs/docker-letsencrypt-nginx-proxy-companion/pull/378/files#diff-8cb95e33acecc83485524b1c4b485120R37) rather than trying to reach the container from the host. That takes care of both Linux / macOS differences and potential future changes in Boulder's Docker Compose file network setup.

PR #379, opened 2018-05-17
Prevent non-Letsencrypt certs removing

SilverFire:
In addition to https://github.com/JrCs/docker-letsencrypt-nginx-proxy-companion/pull/352, I would suggest to keep untouched symlinks to disabled domains, if the certificate does not belong to Letsencrypt.

As a use case example, say I have many `example.com` subdomains and a wildcard certificate `wild.example.com.crt` under `example.com` directory. So I do:

```bash
ln -s example.com/wild.example.com.crt subdomain.example.com.crt
ln -s example.com/wild.example.com.crt otherdomain.example.com.crt
```

and so on. In this case I would not like to see my symlinks removed because they are not created by this proxy companion.

### DIFF
--- a/app/letsencrypt_service
+++ b/app/letsencrypt_service
@@ -90,6 +90,11 @@ function cleanup_links {
     # Return 1 if nothing was removed and 0 otherwise.
     if [[ ${#DISABLED_DOMAINS[@]} -gt 0 ]]; then
       for disabled_domain in "${DISABLED_DOMAINS[@]}"; do
+          certfile="${disabled_domain}.crt"
+          # If certificate is not letsencrypt, don't ever try to remove it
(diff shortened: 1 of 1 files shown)


buchdag:
LGTM

In the longer run we should probably find a more reliable way of determining if a cert is handled by the LE companion than just  "was it issued by Let's Encrypt",  I'm open to suggestions if you have any.

buchdag:
Addition of an empty dotfile next to the certs maybe?

SilverFire:
Thank you for a fast merge 👍 

We can handle it with an index file (in any format) containing domains that were handled by LE, but I think it will bring inconsistency rather than reliability. In my experience I have the following case:

I run new container with `LETSENCRYPT_HOST=example.com`
Companion obtains a LE cert for both domains
Later I get asked to set a cert with higher confirmation level for `example.com`
So I disable LE for this container and set my own cert
As a result, I will probably have index file with information that `example.com` was handled by this companion, but the cert in use will be different. And I don't see how this case can be handled correctly.

Issue #375, opened 2018-05-07
Image latest /app/force_renew does not work Error: nginx-proxy container isn't running.


rene-gomez:
This is my docker-compose.yml 

```
  proxy-companion:
    image: jrcs/letsencrypt-nginx-proxy-companion:latest
    restart: always
    container_name: nginx-letsencrypt
    volumes:
      - /var/www/certs:/etc/nginx/certs:rw
      - /var/run/docker.sock:/var/run/docker.sock:ro
    volumes_from:
      - nginx-proxy

```

The issue appear when try this command: docker exec nginx-letsencrypt /app/force_renew

The console respond:

Error: nginx-proxy container   isn't running.

The result of the docker ps : 
`
jrcs/letsencrypt-nginx-proxy-companion:latest   "/bin/bash /app/en..."   9 minutes ago       Up 9 minutes                                                   nginx-letsencrypt
`

narhen:
I had a similar issue. In my case the labels for nginx and nginx-gen were misconfigured. I used the same label for both. The correct config is to use 
`com.github.jrcs.letsencrypt_nginx_proxy_companion.nginx_proxy` for nginx and `com.github.jrcs.letsencrypt_nginx_proxy_companion.nginx_gen` for nginx-gen

buchdag:
Okay, found the bug, a fix is on the way.

PR #382, opened 2018-05-19
Move the self ID retrieval to /app/functions.sh

buchdag:
This is a fix for #375

This PR move the self ID retrieval to a reusable function, instead of a one time operation that stores the ID in an environment variable at container startup.

The `/app/force_renew` script  being used through `docker exec`, it can't access this environment variable and therefore fails to check for a running nginx container (#321) when using the `--volumes-from` method.

### DIFF
--- a/app/entrypoint.sh
+++ b/app/entrypoint.sh
@@ -8,28 +8,6 @@ if [[ -n "${ACME_TOS_HASH:-}" ]]; then
     echo "simp_le now implicitly agree to the ACME CA ToS."
 fi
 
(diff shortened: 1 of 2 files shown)


buchdag:
It is but `docker exec` can't read this variable because it does not go through the entrypoint script and the variable scope is limited to this same script.

A better approach might be to save and export the result of `get_self_id` in the entrypoint script and do something like `${CONTAINER_ID:$(get_self_id)}` when retrieving the nginx container ID through `--volumes-from`. Sounds okay to you ?

PR #372, opened 2018-05-03
feat: Add dockerfile version for rpi

Javrd:
This PR adds a dockerfile version for rpi from arm32v6 alpine image
It's also use a new ENV var (ARCH_VERSION=armhf) to download correct version of docker-gen.

Related with #316

PR #391, opened 2018-06-04
Use multi-stage build

buchdag:
https://docs.docker.com/develop/develop-images/multistage-build/

This Dockerfile does not fetch a pre built binary of `docker-gen` like the current one but compile it from source during the first stage, then grab the compiled binary for the final image from this stage.

This allows to build the image on any arch the `golang` and `alpine` images are available for from a single Dockerfile.

Tested and working on `arm` 32 bits arch (Raspbian running on  a Raspberry Pi 3).

I feel this is a cleaner and more flexible solution than the usual method of maintaining a different Dockerfile for each arch (as proposed in #372). The drawbacks are:

- requires Docker 17.05+ to build
- the intermediate stage containers have to be manually cleaned

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,27 @@
+FROM golang:1.10-alpine AS go-builder
+
+ENV DOCKER_GEN_VERSION=0.7.4
+
+# Install build dependencies for docker-gen
+RUN apk add --update \
+        curl \
+        gcc \
+        git \
+        make \
+        musl-dev
+
+# Build docker-gen
+RUN go get github.com/jwilder/docker-gen \
+    && cd /go/src/github.com/jwilder/docker-gen \
+    && git checkout $DOCKER_GEN_VERSION \
+    && make get-deps \
+    && make all
+
 FROM alpine:3.7
 
 LABEL maintainer="Yves Blusseau <90z7oey02@sneakemail.com> (@blusseau)"
 
 ENV DEBUG=false \
-    DOCKER_GEN_VERSION=0.7.4 \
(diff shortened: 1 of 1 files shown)


wiget:
Thanks @buchdag for upstreaming this idea. I tested this on HypriotOS/aarch64 on Raspberry Pi 3 B+

rriemann:
@buchdag , do you maintain a fork we can use until your changes are merged?

buchdag:
This PR has already been merged 7 months ago :)

rriemann:
@buchdag , oh my bad - that's true.

Anyway, I still run into a problem when I have the companion as a dependency of Nextcloud on an Scaleway ARM server. I use this docker-compose file: https://github.com/nextcloud/docker/blob/master/.examples/docker-compose/with-nginx-proxy/postgres/fpm/docker-compose.yml

The composer log suggests that arm is not supported:

```
standard_init_linux.go:190: exec user process caused "exec format error"
```

buchdag:
Do you build the image from sources on the Scaleway ARM server ?

rriemann:
I just called `docker-compose build --push` and `docker-compose up -d`.

buchdag:
The goal of this PR was to allow building the image from the Dockerfile on any arch, not providing multi arch image straight from Dockerhub. The image you pull from Dockerhub is still `amd64` only.

Clone the repo on your Scaleway server and build the image from it. It does not work with `docker-compose build` but with docker `build -t jrcs/letsencrypt-nginx-proxy-companion .` from inside the cloned repo (BTW I'm not aware of any `--push` flag to `docker-compose build`).

PR #394, opened 2018-06-05
Use background DH group creation

buchdag:
This PR mirrors what is being done by `jwilder/nginx-proxy` : create the DH group in the background with a low priority job and use a temporary pre-generated 2048 bits DH group while it's being done. This way the container can start working immediately while still being secure. 

Credits goes to @kamermans for the original code contibution to `jwilder/nginx-proxy`.


buchdag:
Might fix #396

app/entrypoint.sh
```diff
@@ -42,17 +42,53 @@ function check_writable_directory {
...
-        openssl dhparam -out /etc/nginx/certs/.dhparam.pem.tmp $DHPARAM_BITS
-        mv /etc/nginx/certs/.dhparam.pem.tmp /etc/nginx/certs/dhparam.pem || exit 1
+
+    # If a dhparam file is not available, use the pre-generated one and generate a new one in the background.
+    local PREGEN_DHPARAM_FILE="/app/dhparam.pem.default"
+    local DHPARAM_FILE="/etc/nginx/certs/dhparam.pem"
+    local GEN_LOCKFILE="/tmp/le_companion_dhparam_generating.lock"
+
+    # The hash of the pregenerated dhparam file is used to check if the pregen dhparam is already in use
+    local PREGEN_HASH=$(md5sum "$PREGEN_DHPARAM_FILE" | cut -d ' ' -f1)
```
NicolasDorier: why md5sum and not sha256sum ?

test/setup/setup-nginx-proxy.sh
```diff
@@ -7,6 +7,7 @@ case $SETUP in
   2containers)
     docker run -d -p 80:80 -p 443:443 \
       --name $NGINX_CONTAINER_NAME \
+      --env "DHPARAM_BITS=256" \
```
NicolasDorier: 256 is insecure

app/entrypoint.sh
buchdag: Because we're just doing a quick internal file comparison and an md5 hash is more than enough for that purpose.

Worst case and extremely improbable scenario is you generate a dh params file whose md5 collide with the pre generated dh params and a new one is generated again at container restart.

NicolasDorier: `sha256sum` is 2 letters longer.

Also unrelated, if I understand using this new PR means that for 1 deployment there will be 2 certificate requests right? It need to be documented so people don't go over the limits for their domain unexpectedly.

buchdag: No, this is completely unrelated to certificate requests.

What this PR does is outlined in the first message:

- check if a custom (ie different from the bundled one) DH parameters file is present.
- if not, use the pre generated DH parameters file bundled with the image and start generating a new one in the background, with a low priority process.
- when the new DH parameters file is generated, reload nginx so it start using it.

The goal here was to remove the initial wait and service unavailability due to foreground DH parameters file generation (which could be quite long on some low end hardware) whilst still using a strong 2048 bits DH parameters.

NicolasDorier: Awesome, I thought the DH parameter was used for the private key generation of the certificate, but it is used only in the diffie hellman.

PR #400, opened 2018-06-12
Remove old procps specific pkill command

buchdag:
We don't use procps anymore.


buchdag:
#399 will have to be merged first so the test can pass

PR #401, opened 2018-06-14
invoke /app/docker-entrypoint.sh when reloading nginx-proxy

thomseddon:
The /app/docker-entrypoint.sh within nginx-proxy [sets the RESOLVERS environment variable](https://github.com/jwilder/nginx-proxy/blob/master/docker-entrypoint.sh#L22), so when docker-proxy is invoked without it the resolvers aren't set in the nginx config.

The more general problem, which this PR doesn't fix, is that all other environment variables are also clobbered when the config is generated by these means.
Unfortunately, as docker-gen is run via forego within docker-proxy, a HUP signal sent to the docker-proxy container won't be passed through (it just kills the container). I think the actual fix here is probably to change the handling of HUP's within forego?

Might also be worth a note in the README as it took me ages to work out why this environment variable seemed not to be set sometimes...

### DIFF
--- a/app/functions.sh
+++ b/app/functions.sh
@@ -179,7 +179,7 @@ function reload_nginx {
         if [[ -n "${_nginx_proxy_container:-}" ]]; then
             echo "Reloading nginx proxy (${_nginx_proxy_container})..."
             docker_exec "${_nginx_proxy_container}" \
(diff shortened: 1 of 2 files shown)


buchdag:
The `docker_api` test unit has to be modified to take your changes into account.

thomseddon:
:) So I see! Just fixing now.

thomseddon:
Done :)

buchdag:
LGTM, thanks for this PR !

PR #403, opened 2018-06-14
Remove set -x for debug

buchdag:
In my opinion `set -x` makes the logs quite unreadable and is not of much help when trying to troubleshoot an issue. I'd rather remove it and work on cleaner and clearer debug logging.

### DIFF
--- a/app/entrypoint.sh
+++ b/app/entrypoint.sh
@@ -93,8 +93,6 @@ is being created."
 
 source /app/functions.sh
 
-[[ $DEBUG == true ]] && set -x
-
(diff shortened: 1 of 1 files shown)


buchdag:
@JrCs would you prefer to keep `set -x` when `DEBUG` is set, and if yes why ?

JrCs:
No it's ok for me, if we have usefull debug trace.
Perhaps a DEV variable to allow full bash debug (with set-x) if you think it can be usefull for us.

buchdag:
I'm going to merge this as is for now and add another way to allow `set -x` later (soon), right now it's really making my life harder more than anything.

PR #408, opened 2018-06-18
Disable example.com notification emails

NicolasDorier:
Let's encrypt now consider `example.com` emails invalid. This is a breaking change as lot's of people were using it before as a dummy.
This PR makes sure that we don't try to use `example.com` as a notification email.

Note that https://tools.ietf.org/html/rfc2606 also define that `example.com` should be only used in example and not as a dummy.

### DIFF
--- a/app/letsencrypt_service
+++ b/app/letsencrypt_service
@@ -144,7 +144,7 @@ function update_certs {
 
         email_varname="LETSENCRYPT_${cid}_EMAIL"
         email_address="${!email_varname}"
-        if [[ "$email_address" != "<no value>" ]]; then
+        if [[ "$email_address" != "<no value>" ]] || [[ $email_address == *@example.com ]]; then
             params_d_str+=" --email $email_address"
         fi
 


buchdag:
> Let's encrypt now consider `example.com` emails invalid. This is a breaking change as lot's of people were using it before as a dummy.

The doc [clearly state that the address **must be valid**](https://github.com/JrCs/docker-letsencrypt-nginx-proxy-companion#contact-address) and the email has been made entirely optional more than six months ago.

I think that if a check for `@example.com` addresses has to be made somewhere, it's in the ACME client, not in this container code. This has been [suggested](https://github.com/JrCs/docker-letsencrypt-nginx-proxy-companion/issues/263#issuecomment-334252820) and [done](https://github.com/zenhack/simp_le/issues/76) before on a similar issue.

@JrCs considering the above, I'm in favor of closing this PR. Your thought ?

JrCs:
@buchdag i'm agree with you.

NicolasDorier:
> The doc clearly state that the address must be valid and the email has been made entirely optional more than six months ago.

It is not optional, you need to set it equals to `<no value>` to not use a mail, which undocumented and not really user friendly. Omitting the LETSENCRYPT_EMAIL does not work.

NicolasDorier:
Yes I tried it, because it took me ages to find out.

Here is [my docker-compose](https://github.com/btcpayserver/btcpayserver-docker/blob/master/docker-compose-generator/docker-fragments/btcpayserver.yml#L26).

Removing `:-<no value>` result in warning from docker-compose because the variable is not set. I tried several other solution I can't remember now, but found out the only solution was what I changed to.

buchdag:
Ok, I get your issue. Your use case is a bit specific as you set a `LETSENCRYPT_EMAIL` variable on your container no matter what whereas most people will either set this variable or not set it at all. That was the intended use case when we made `LETSENCRYPT_EMAIL` optional, and work as it should.

I've got change that should take care of this use case (negating the need for `:-<no value>`) queued as part of another feature in `dev`, would it be useful to you if I merged them ?

NicolasDorier:
No rush for now as what I did work. That said, I would be happy to know when this get merged so I cleanup my yml.

Issue #420, opened 2018-08-02
Every unknown https domain/subdomain redirected to the first active container ?


dridri:
Looks like there is some default feature that should not be default.
This behavior only happens with HTTPS (HTTP works correctly) : every unknown host is proxied to the first active container.

For example considering a container with VHOST=sub1.domain.com : opening domain.com / sub2.domain.com / *.domain.com in my web browser should give me a server-not-found error.
But it actually throws me on the first container available, in this case the one with sub1.domain.com, hence when I browse to sub2.domain.com, I end up in index page of sub1.domain.com.

This also happens directly with domains : if domain2.com is on the same machine as domain.com, and no domain2.com container is active, every https calls end up on the sub1.domain.com

This does not look very secure to me.
(Actually, I don't know if it's directly nginx related or letsencrypt related, so I post this issue here first)

dridri:
Looks like I'm duplicating existing issues, closing

buchdag:
@dridri the discussion is actually on #411

Issue #416, opened 2018-07-16
503 for default or non-existent virtual hosts


mcnesium:
There should be a 503 or something, if a request comes in on a domain or IP address that is not defined by `VIRTUAL_HOST` env var. Instead, it just redirects to the first defined one.

According to [this](https://serverfault.com/a/329953) there must be a `default` added to the default `listen`-directive, for which I assume [here is the right place](https://github.com/jwilder/nginx-proxy/blob/master/nginx.tmpl#L94). This file gets re-generated each time the proxy container starts though.

Is this within the scope of this project or rather upstream at [jwilder/nginx-proxy](https://github.com/jwilder/nginx-proxy)?

buchdag:
According to the answer from server fault that you quoted:

> If you omit the default parameter, the first server block is treated as the default.

So [this](https://github.com/jwilder/nginx-proxy/blob/master/nginx.tmpl#L94) `server` block is already the default.

I think you issue is instead yet another duplicate of #74, #373 and #411 

If it is, could you close this issue and continue the discussion on #411 ?

mcnesium:
Looks like it. My experience with nginx is very little, I can only compare it to Apache's `000-default.conf` that prevents requests to arbitrary domains/IP-addresses to reach a wrong web service.

Issue #411, opened 2018-07-04
Undefined routes leaking random certs.


vicary:
# Current behavior
When visiting the proxy with an undefined domain, in my case the domain is directed with a wildcard `A` record, while having at least one valid upstream container with `LETSENCRYPT_HOST` defined, the current nginx conf generated will serve the first cert available in the conf file.

This is essentially leaking one of the upstream domains in an unpredictable way.

# Expected behavior
It should fallback to `http` and shows 503, the same behavior as if `nginx-proxy` working without `letsencrypt-nginx-proxy-companion`.

buchdag:
Please see #373

buchdag:
I feel that the issue is more on the `nginx-proxy` side if the config generated by `docker-gen` requires a default.cert and default.key files and does not work as expected when they are not present.

Let me know your advice on this.

buchdag:
Not at the moment but I'm considering it.

buchdag:
Ok, judging by the number of people that reported this as an issue the past few weeks, I think automatic creation of this default certificate should be added to the container. I'm already working on it.

buchdag:
Update : I'm still working on this, just not very fast 'cause summer holidays. I intend this to be fixed by the end of august at the very latest.

Meanwhile, you can manually create a default cert and key with this command, just replace `your-companion-container` with the actual name or id or your **letsencrypt-nginx-proxy-companion** container:

```
docker exec your-companion-container openssl req -x509 \
    -newkey rsa:4096 -sha256 -nodes -days 365 \
    -subj "/CN=default" \
    -keyout /etc/nginx/certs/default.key \
    -out /etc/nginx/certs/default.crt
```

buchdag:
This should be fixed by #423

vicary:
Guess I'll have to do more searching, closing this issue. Thanks again!

Issue #373, opened 2018-05-05
companion bugs default https host for nginx proxy


ipepe:
Hi. I noticed that when using companion (separate containers scenario) there are no `default.key` and `default.crt` and that makes `nginx.tmpl` to not generate a default (and first) 'server' in configuration that would respond with 503. Because of that, when creating such stack, anyone can query my server on https by ip and nginx will respond with first valid https configuration, and not 503 like it should. How can we fix this?

buchdag:
Hi.

Correct me if I'm wrong, but I think the same thing would happen even if one does not use `letsencrypt-nginx-proxy-companion` at all and just manually put certificates on the `/etc/nginx/certs` folder as described in the `nginx-proxy` doc.

ipepe:
Yeah, I think You are right. Today I figured out how to setup this correctly, although it would be nice to have this information in readme or somewhere.

So basically I change my claim on this issue from "Bug" to "Needs more information in Readme"

vicary:
I guess my issue #411 is describing the same thing, but I am not sure if we should simply document this and call it a day.

Making up some invalid placeholder certs for default doesn't quite make sense in production.

I see that users of `nginx-proxy` is very likely to start without a valid default upstream, and may never will.

How about disabling access to default server in the default template for `nginx-proxy` and rely solely on `VIRTUAL_HOST`? Users do need a default upstream can always override it with a separate `docker-gen` container.

PR #423, opened 2018-08-11
Automatic creation of default cert and private key

buchdag:
This is a proposed fix for #411 (and the related #74 #373 #416 and #420).

This PR check at container startup if both `default.crt` and `default.key` are present in `/etc/nginx/certs` and create them otherwise. This auto-created certificate is self signed, has subject CN=letsencrypt-nginx-proxy-companion and is valid for 365 days.

If the startup function detect a user provided default certificate and key (ie a certificate which subject does not match `letsencrypt-nginx-proxy-companion`), it won't overwrite it.

If the startup function detect an auto-created certificate which remaining validity is less than three months, it will re-create a new one.

Some test units and functions had to be modified / expanded to take this new feature into account.

This feature comes with its own test unit.


buchdag:
Under which section of the doc do you think this should go ?

JrCs:
Feel free to choose. Just a little sentence to said that we can use default self-signed certificates.

buchdag:
Travis is having network related issues since two days ago, might be a while before the tests finally succeed.

Issue #412, opened 2018-07-04
Permissions: private keys readable by all users on host machine


dlazerka:
I noticed that certificates together with key.pem are stored on host machine with `-rw-r--r--` permissions.
I believe that should be `-rw-------`.

Just checked how LetsEncrypt certbot does it -- `-rw-------` and `drwx------` on containing directory.

buchdag:
relevant certbot reference:

```
[drwxr-xr-x]  /etc/letsencrypt
├── [drwx------]  accounts
│   └── [drwx------]  acme-v01.api.letsencrypt.org
│       └── [drwx------]  directory
│           └── [drwx------]  somehash
│               ├── [-rw-r--r--]  meta.json
│               ├── [-r--------]  private_key.json
│               └── [-rw-r--r--]  regr.json
├── [drwx------]  archive
│   ├── [drwxr-xr-x]  domain.tld
│   │   ├── [-rw-r--r--]  cert1.pem
│   │   ├── [-rw-r--r--]  chain1.pem
│   │   ├── [-rw-r--r--]  fullchain1.pem
│   │   └── [-rw-r--r--]  privkey1.pem
│   └── [drwxr-xr-x]  subdomain.domain.tld
│       ├── [-rw-r--r--]  cert1.pem
│       ├── [-rw-r--r--]  chain1.pem
│       ├── [-rw-r--r--]  fullchain1.pem
│       └── [-rw-r--r--]  privkey1.pem
└──[drwx------]  live
    ├── [drwxr-xr-x]  domain.tld
    │   ├── [lrwxrwxrwx]  cert.pem -> ../../archive/domain.tld/cert1.pem
    │   ├── [lrwxrwxrwx]  chain.pem -> ../../archive/domain.tld/chain1.pem
    │   ├── [lrwxrwxrwx]  fullchain.pem -> ../../archive/domain.tld/fullchain1.pem
    │   └── [lrwxrwxrwx]  privkey.pem -> ../../archive/domain.tld/privkey1.pem
    └── [drwxr-xr-x]  subdomain.domain.tld
        ├── [lrwxrwxrwx]  cert.pem -> ../../archive/subdomain.domain.tld/cert1.pem
        ├── [lrwxrwxrwx]  chain.pem -> ../../archive/subdomain.domain.tld/chain1.pem
        ├── [lrwxrwxrwx]  fullchain.pem -> ../../archive/subdomain.domain.tld/fullchain1.pem
[...]

PR #436, opened 2018-09-04
Restrain permissions on sensitive files

buchdag:
This PR is a follow up to #412. It add checks and adjustment of ownership / permissions of sensitive files (private keys and ACME account keys) and enclosing folders + corresponding test unit.

The following files and folders are are currently adjusted by the PR:

```
[drwx------]  /etc/nginx/certs
├── [-rw-------]  default.key
├── [drwx------]  accounts
│   └── [drwx------]  acme-v01.api.letsencrypt.org
│       └── [drwx------]  directory
│           └── [-rw-------]  default.json
├── [drwx------]  domain.tld
│   └── [-rw-------]  key.pem
└── [drwx------]  subdomain.domain.tld
    └── [-rw-------]  key.pem
```

I'm not sure yet about changing /etc/nginx/certs ownership and permissions, I suspect it might break things for people using host volumes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,6 +48,7 @@ after_failure:
   - docker logs certs_single
   - docker logs certs_san
   - docker logs force_renew
+  - docker logs permissions
   - docker logs symlinks
   - docker logs boulder
   - if [[ $SETUP = "3containers" ]]; then docker logs $DOCKER_GEN_CONTAINER_NAME; fi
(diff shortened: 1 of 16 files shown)


buchdag:
@JrCs any advice on changing `/etc/nginx/certs` permissions vs leaving them as they are ?

It works in both case with docker volumes, but it might be an inconvenience for people using host volumes (ie `-v /path/to/certs:/etc/nginx/certs` ).

buchdag:
I've decided not to touch `/etc/nginx.certs` ownership and permissions.

@JrCs can you review and approve ?

buchdag:
ping @JrCs

PR #442, opened 2018-09-24
New environmental variable LETSENCRYPT_RESTART_CONTAINER

### LINKED ISSUE #434: Restart Containers on Cert Renewal
SCENARIO:
Let's say you have a micro-service that doesn't only serve HTTP traffic, but also other kind of traffic (e.g. FTP).
Now let's say that we do not only want to encrypt the web traffic, but also the ftp traffic (using FTPS). This can be done easily by integrating the keys generated by the letsencrypt-companion into the FTP container and use them for FTPS.

The problem now arises on Cert renewal. While the letsencrypt-companion restart the nginx reverse-proxy container (meaning that the web traffic uses the new certificates), the FTP container still uses the old ones and also needs to be restarted.

SUGGESTION:
Add an extra environmental variable (e.g. LE_RESTART), that if set true, signals the letsencrypt-companion to restart the respective container.
To avoid overheads, it would also make sense to additionally check the LETSENCRYPT_HOST and only restart if the respective cert of the container is renewed.

Greek64:
This PR adds an additional functionality via an additional environmental variable.

Containers with the `LETSENCRYPT_RESTART_CONTAINER` variable set to true will be restarted when their respective certificates are updated/modified.

Fixes #434

### DIFF
--- a/app/letsencrypt_service
+++ b/app/letsencrypt_service
@@ -126,9 +126,10 @@ function update_certs {
     unset LETSENCRYPT_CONTAINERS
     # shellcheck source=/dev/null
     source /app/letsencrypt_service_data
-
+    
     should_reload_nginx='false'
     for cid in "${LETSENCRYPT_CONTAINERS[@]}"; do
+        should_restart_container='false'
         # Derive host and email variable names
         host_varname="LETSENCRYPT_${cid}_HOST"
         # Array variable indirection hack: http://stackoverflow.com/a/25880676/350221
@@ -150,7 +151,7 @@ function update_certs {
         if [[ "$cert_keysize" == "<no value>" ]]; then
             cert_keysize=$DEFAULT_KEY_SIZE
         fi
-
+        
         test_certificate_varname="LETSENCRYPT_${cid}_TEST"
         le_staging_uri="https://acme-staging.api.letsencrypt.org/directory"
         if [[ $(lc "${!test_certificate_varname:-}") == true ]] || \
@@ -231,9 +232,9 @@ function update_certs {
         if [[ $simp_le_return -ne 2 ]]; then
           for domain in "${!hosts_array}"; do
             if [[ "$acme_ca_uri" == "$le_staging_uri" ]]; then
-              create_links "_test_$base_domain" "$domain" && should_reload_nginx='true'
+              create_links "_test_$base_domain" "$domain" && should_reload_nginx='true' && should_restart_container='true'
             else
-              create_links "$base_domain" "$domain" && should_reload_nginx='true'
+              create_links "$base_domain" "$domain" && should_reload_nginx='true' && should_restart_container='true'
             fi
             touch "${certificate_dir}/.companion"
             set_ownership_and_permissions "${certificate_dir}/.companion"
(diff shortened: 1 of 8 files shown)


buchdag:
Hi. I haven't forgotten about this PR but I'm still unsure about merging it (because of the project scope issue).

By the way if this is to be merged it'll require a corresponding test unit.

Greek64:
Yeah, don't worry.  
As said, I will try to keep the fork updated for the guys that want this functionality as long as this isn't merged (which may be never).

I didn't think of the test unit.  
I will look into that when I have time.

buchdag:
No hurry for the test unit, at least until we've reached a decision about merging the feature or not.

Greek64:
Another thing that I noticed in the test units:
Shouldn't the loop in  [line 34](https://github.com/JrCs/docker-letsencrypt-nginx-proxy-companion/blob/f936be75a3a4a53e89249f9b4e353c31ec497351/test/tests/default_cert/run.sh#L34) in the default_certs test unit contain a `break`?

buchdag:
> Another thing that I noticed in the test units:
> Shouldn't the loop in [line 34](https://github.com/JrCs/docker-letsencrypt-nginx-proxy-companion/blob/f936be75a3a4a53e89249f9b4e353c31ec497351/test/tests/default_cert/run.sh#L34) in the default_certs test unit contain a `break`?

Yes it should, thanks for spotting that.

I've been lacking spare time for the last few weeks, I'll come back to this PR as soon as I can. Got a few thinks I want to clear first.

Greek64:
> The thing is, that the test units on master fail on my system (even without me changing anything), and so I am reluctant to push them on these PR branch, until they run on my system.

I found the reason why the test units fail locally. 
See #482

Greek64:
I updated the PR to the newest master commit and pushed the new test units.
Now everything passes.

buchdag:
Okay, I've just reviewed your PR, thanks for your work.

I think a separate test unit for this feature would be more fitting, rather than adding more logic to three existing tests. You could just take the basic `cert_single` test, remove all the cert verification code and just check that the container is restarted correctly.

Also if your test generate extra files (`docker_event_out.txt`), don't forget to add them to `.gitignore`, just in case.

buchdag:
BTW if this wasn't clear this PR will be merged as soon as the last details are ironed out 👍

Greek64:
Added a separate "container_restart" test unit, and added temp file to .gitignore.

app/letsencrypt_service
```diff
@@ -126,9 +126,10 @@ function update_certs {
     unset LETSENCRYPT_CONTAINERS
     # shellcheck source=/dev/null
     source /app/letsencrypt_service_data
-
+
```
buchdag: Whitespace issues ?

app/letsencrypt_service
```diff
@@ -150,7 +151,7 @@ function update_certs {
         if [[ "$cert_keysize" == "<no value>" ]]; then
             cert_keysize=$DEFAULT_KEY_SIZE
         fi
-
+
```
buchdag: Same here.

buchdag:
If you can cleanup the two remaining whitespace typos and write a line of doc explaining this feature I think we're good to go.

app/letsencrypt_service
Greek64: Probably a side effect of "automatic indentation" of my Editor.

app/letsencrypt_service
Greek64: Same as above

Greek64:
> If you can cleanup the two remaining whitespace typos and write a line of doc explaining this feature I think we're good to go.

Done

Issue #465, opened 2018-11-12
 letsencrypt-nginx-proxy-companion is removing out of its context symlinks in /etc/nginx/certs


ejbp:
I'm not able to have a mixed configuration with domains that are under letsencrypt-nginx-proxy-companion management and domains that are not, having in mind that jwilder/nginx-proxy is the same and using the same certs dir for all domains. What's happening is that letsencrypt-nginx-proxy-companion is removing my custom certificates since it considers that they are unsed.

log:
```
letsencrypt-nginx-proxy-companion    | Symlinked domains: yyy.com xxx.com
letsencrypt-nginx-proxy-companion    | Enabled domains: xxx.com
letsencrypt-nginx-proxy-companion    | Disabled domains: yyy.com
letsencrypt-nginx-proxy-companion    | Some domains are disabled. Check them to remove unused symlinks.
letsencrypt-nginx-proxy-companion    | 
letsencrypt-nginx-proxy-companion    | Checking domain yyy.com: 
letsencrypt-nginx-proxy-companion    | Checking yyy.com.crt - removing.
letsencrypt-nginx-proxy-companion    | Checking yyy.com.key - removing.
letsencrypt-nginx-proxy-companion    | Checking yyy.com.dhparam.pem
letsencrypt-nginx-proxy-companion    | Checking yyy.com.chain.pem - removing.
letsencrypt-nginx-proxy-companion    | Unused domains checking is finished.
```

docker-compose.yml:

```
services:
  letsencrypt_nginx_proxy_companion:
    image: jrcs/letsencrypt-nginx-proxy-companion
    container_name: letsencrypt-nginx-proxy-companion
    volumes_from:
      - load_balancer
    volumes:
      - "/var/run/docker.sock:/var/run/docker.sock:ro"
      - "${LETSENCRYPT_CERTS_PATH}:/etc/nginx/certs:rw"
    environment:
      - NGINX_DOCKER_GEN_CONTAINER=load_balancer
  load_balancer:
    image: jwilder/nginx-proxy
    container_name: load_balancer
    ports:
      - 80:80
      - 443:443
    volumes:
      - ./docker/load-balancer/nginx-proxy-vhost.d:/etc/nginx/vhost.d
      - /var/run/docker.sock:/tmp/docker.sock:ro
      - /usr/share/nginx/html
      - "${LETSENCRYPT_CERTS_PATH:/etc/nginx/certs:ro"
    labels:
[...]

PR #472, opened 2018-11-23
Follow certificates managed by the companion with a dot file

buchdag:
As discussed in #465, this PR should fix the remaining issues with hybrid configurations (certs obtained through the companion stored and used along certs obtained another way).

> [...] include a dot file next to the certs handled by the companion during creation / renewal (most importantly before the symlink cleanup function). Then the symlink cleanup function check for the presence of this dot file. If it's not there, those files are not handled by the companion, don't touch the symlinks.

@ejbp any thought before merging this ?


ejbp:
@buchdag It looks very nice. I've nothing further to add. Thanks for the update 👍

PR #480, opened 2018-12-13
Make the DEBUG env var case insensitive

buchdag:
The DEBUG env var is actually case sensitive (eg it will work when set to `true` but not to `True` as it should). This PR makes it case insensitive.


desimaniac:
Thanks!

PR #485, opened 2018-12-22
Add LETSENCRYPT_MIN_VALIDITY variable

### LINKED ISSUE #477: Automatic Renewal
The le-companion is renewing the Certificates when they are valid for less than 30 days.
This is currently achieved by simply not setting the `--valid_min` option of the `simp_le` client (i.e. using the default setting).

Even if this is documented in the Manifesto of `simp_le` (https://github.com/zenhack/simp_le), we are relying on a default value, about something we explicitly want to have. In others words it would make more sense to set this explicitly.

Actually we could add another environmental variable (e.g `LETSENCRYPT_MIN_VALID`), so that every user can customize this to their liking (of course with the according sanity checks, to prevent a min-validity larger that LE will give)

And also, as it currently stands, the auto renewal process is not even tested by the test units (only the default_cert unit tests the renewal of the default certificate).
Adding a variable would allow to "spoof" the validity check (by giving a higher min validity than the current certs have), in order to test the auto renewal feature.

If interested, I could make a PR for this (including new test unit for auto renewal test)

Greek64:
This PR adds another environment variable called `LETSENCRYPT_MIN_VALIDITY`, that can be used to set the minimum validity of certificates per container.

Closes #477

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,7 +166,10 @@ If you want to create multi-domain ([SAN](https://www.digicert.com/subject-alter
 If you want to create test certificates that don't have the 5 certs/week/domain limits define the `LETSENCRYPT_TEST` environment variable with a value of `true` (in the containers where you request certificates with `LETSENCRYPT_HOST`). If you want to do this globally for all containers, set `ACME_CA_URI` as described below.
 
 ##### Automatic certificate renewal
-Every hour (3600 seconds) the certificates are checked and every certificate that will expire in the next [30 days](https://github.com/kuba/simp_le/blob/ecf4290c4f7863bb5427b50cdd78bc3a5df79176/simp_le.py#L72) (90 days / 3) are renewed.
+Every hour (3600 seconds) the certificates are checked and per default every certificate that will expire in the next [30 days](https://github.com/zenhack/simp_le/blob/a8a8013c097910f8f3cce046f1077b41b745673b/simp_le.py#L73) (90 days / 3) is renewed.
+
+If you want to manually set a different minimum validity for certificates, you can set the `LETSENCRYPT_MIN_VALIDIDTY` environment variable (in each container that defines the `LETSENCRYPT_HOST` variable) to the desired period in seconds.  
+Note that the possible values are internally capped at an upper bound of 7603200 (88 days) and a lower bound of 7200 (2 hours) as a security margin, considering that the Let's Encrypt CA does only issues certificates with a lifetime of [90 days](https://letsencrypt.org/2015/11/09/why-90-days.html) (upper bound), the rate limits imposed on certificate renewals are [5 per week](https://letsencrypt.org/docs/rate-limits/) (upper bound), and the fact that the certificates are checked and renewed accordingly every hour (lower bound).
 
 ##### Force certificates renewal
 If needed, you can force a running letsencrypt-nginx-proxy-companion container to renew all certificates that are currently in use. Replace `nginx-letsencrypt` with the name of your letsencrypt-nginx-proxy-companion container in the following command:
--- a/app/letsencrypt_service
+++ b/app/letsencrypt_service
@@ -8,6 +8,8 @@ ACME_CA_URI="${ACME_CA_URI:-https://acme-v01.api.letsencrypt.org/directory}"
 DEFAULT_KEY_SIZE=4096
 REUSE_ACCOUNT_KEYS="$(lc ${REUSE_ACCOUNT_KEYS:-true})"
 REUSE_PRIVATE_KEYS="$(lc ${REUSE_PRIVATE_KEYS:-false})"
+MIN_VALIDITY_CAP=7603200
+DEFAULT_MIN_VALIDITY=2592000
 
 function create_link {
     local -r source=${1?missing source argument}
@@ -174,7 +176,28 @@ function update_certs {
 
         [[ "$(lc $DEBUG)" == true ]] && params_d_str+=" -v"
         [[ $REUSE_PRIVATE_KEYS == true ]] && params_d_str+=" --reuse_key"
-        [[ "${1}" == "--force-renew" ]] && params_d_str+=" --valid_min 7776000"
+
+        min_validity="LETSENCRYPT_${cid}_MIN_VALIDITY"
+        min_validity="${!min_validity}"
+        if [[ "$min_validity" == "<no value>" ]]; then
+            min_validity=$DEFAULT_MIN_VALIDITY
+        fi
+        # Sanity Check
+        # Upper Bound
+        if [[ $min_validity -gt $MIN_VALIDITY_CAP ]]; then
+            min_validity=$MIN_VALIDITY_CAP
+        fi
+        # Lower Bound
+        if [[ $min_validity -lt $(($seconds_to_wait * 2)) ]]; then
+            min_validity=$(($seconds_to_wait * 2))
+        fi
+
+        if [[ "${1}" == "--force-renew" ]]; then
+            # Manually set to highest certificate lifetime given by LE CA
+            params_d_str+=" --valid_min 7776000"
+        else
+            params_d_str+=" --valid_min $min_validity"
+        fi
 
         # Create directory for the first domain,
         # make it root readable only and make it the cwd
--- a/app/letsencrypt_service_data.tmpl
+++ b/app/letsencrypt_service_data.tmpl
@@ -11,6 +11,7 @@ LETSENCRYPT_{{ $cid }}_KEYSIZE="{{ $container.Env.LETSENCRYPT_KEYSIZE }}"
 LETSENCRYPT_{{ $cid }}_TEST="{{ $container.Env.LETSENCRYPT_TEST }}"
 LETSENCRYPT_{{ $cid }}_ACCOUNT_ALIAS="{{ $container.Env.LETSENCRYPT_ACCOUNT_ALIAS }}"
 LETSENCRYPT_{{ $cid }}_RESTART_CONTAINER="{{ $container.Env.LETSENCRYPT_RESTART_CONTAINER }}"
+LETSENCRYPT_{{ $cid }}_MIN_VALIDITY="{{ $container.Env.LETSENCRYPT_MIN_VALIDITY }}"
 {{ end }}
 
 {{ end }}
--- a/test/config.sh
+++ b/test/config.sh
@@ -12,6 +12,7 @@ imageTests+=(
 	certs_single
 	certs_san
 	force_renew
+	certs_validity
 	container_restart
 	permissions_default
 	permissions_custom
--- a/test/setup/setup-boulder.sh
+++ b/test/setup/setup-boulder.sh
@@ -11,12 +11,22 @@ setup_boulder() {
       $GOPATH/src/github.com/letsencrypt/boulder
   pushd $GOPATH/src/github.com/letsencrypt/boulder
   if [[ "$(uname)" == 'Darwin' ]]; then
+    # Set Standard Ports
     sed -i '' 's/ 5002/ 80/g' test/config/va.json
     sed -i '' 's/ 5001/ 443/g' test/config/va.json
+    # Set certificate lifetime to 88 days
+    sed -i '' 's/2160h/2112h/g' test/config/ca-a.json
+    sed -i '' 's/2160h/2112h/g' test/config/ca-b.json
+    # Modify custom rate limit
     sed -i '' 's/le.wtf,le1.wtf/le1.wtf,le2.wtf,le3.wtf/g' test/rate-limit-policies.yml
   else
+    # Set Standard Ports
     sed --in-place 's/ 5002/ 80/g' test/config/va.json
     sed --in-place 's/ 5001/ 443/g' test/config/va.json
+    # Set certificate lifetime to 88 days
+    sed --in-place 's/2160h/2112h/g' test/config/ca-a.json
+    sed --in-place 's/2160h/2112h/g' test/config/ca-b.json
+    # Modify custom rate limit
     sed --in-place 's/le.wtf,le1.wtf/le1.wtf,le2.wtf,le3.wtf/g' test/rate-limit-policies.yml
   fi
   docker-compose build --pull
--- a/test/tests/certs_validity/expected-std-out.txt
+++ b/test/tests/certs_validity/expected-std-out.txt
@@ -1,0 +1,13 @@
+Started letsencrypt container for test certs_validity
+Started test web server for le1.wtf
+Started test web server for le2.wtf
+Started test web server for le3.wtf
+Symlink to le1.wtf certificate has been generated.
+The link is pointing to the file ./le1.wtf/fullchain.pem
+Symlink to le2.wtf certificate has been generated.
+The link is pointing to the file ./le2.wtf/fullchain.pem
+Symlink to le3.wtf certificate has been generated.
+The link is pointing to the file ./le3.wtf/fullchain.pem
+Certificate for le1.wtf was not renewed.
+Certificate for le2.wtf was not renewed.
+Certificate for le3.wtf was renewed.
--- a/test/tests/certs_validity/run.sh
+++ b/test/tests/certs_validity/run.sh
@@ -1,0 +1,93 @@
+#!/bin/bash
+
+## Test for the LETSENCRYPT_MIN_VALIDITY environment variable.
+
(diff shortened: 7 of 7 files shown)


buchdag:
`LETSENCRYPT_MIN_VALIDITY` should be capped a bit lower than 90 days, because capping at 90 days will have the same end result as no capping at all: the container will attempt to renew the certificate on every run of the service loop and will hit LE rate limits pretty fast.

Considering LE rate limiting on this is 5 renewal / week for a given cert, 7655040 seconds (7776000 - ( ( 24 * 7 * 60 * 60 ) / 5 ) seconds seems to be a more sensible capping.

88 days / 7603200 (ie taking a small security margin over LE rate limit and rounding it to the nearest whole day) seconds would also make sense .

README.md
```diff
@@ -166,7 +166,11 @@ If you want to create multi-domain ([SAN](https://www.digicert.com/subject-alter
 If you want to create test certificates that don't have the 5 certs/week/domain limits define the `LETSENCRYPT_TEST` environment variable with a value of `true` (in the containers where you request certificates with `LETSENCRYPT_HOST`). If you want to do this globally for all containers, set `ACME_CA_URI` as described below.
 
 ##### Automatic certificate renewal
-Every hour (3600 seconds) the certificates are checked and every certificate that will expire in the next [30 days](https://github.com/kuba/simp_le/blob/ecf4290c4f7863bb5427b50cdd78bc3a5df79176/simp_le.py#L72) (90 days / 3) are renewed.
+Every hour (3600 seconds) the certificates are checked and per default every certificate that will expire in the next [30 days](https://github.com/kuba/simp_le/blob/ecf4290c4f7863bb5427b50cdd78bc3a5df79176/simp_le.py#L72) (90 days / 3) is renewed.
```
buchdag: While we are changing this line, could you update the link URL to point to the actual version / repo of `simp_le` we are using ?

https://github.com/zenhack/simp_le/blob/a8a8013c097910f8f3cce046f1077b41b745673b/simp_le.py#L73

Greek64:
> `LETSENCRYPT_MIN_VALIDITY` should be capped a bit lower than 90 days, because capping at 90 days will have the same end result as no capping at all: the container will attempt to renew the certificate on every run of the service loop and will hit LE rate limits pretty fast.

Yeah, originally I wanted to cap it to a lower value for the stated reasons, but then comes the problem of testing.  
How can the test unit test this feature by "simulating" a passing of 48 hours in a reasonable time?

> Considering LE rate limiting on this is 5 renewal / week for a given cert

According to their [site](https://letsencrypt.org/docs/rate-limits/), what you are referring to are the **Duplicate Certificate** rate limits. But they also have a **Renewal Exemption** (which applies to Duplicate Certificates).
I quote:

>To make sure you can always renew your certificates when you need to, we have a Renewal Exemption to the Certificates per Registered Domain limit. Even if you’ve hit the limit for the week, you can still issue new certificates that count as renewals.

I also originally wanted to add also a lower bound, to prevent unwanted downtime due to misconfiguration, since the check is triggered hourly and the user could give a period smaller than 3600 seconds.

buchdag:
My understanding is that Duplicate and Renewal are considered to be the same from a rate limit perspective:

> This is the same definition used for the Duplicate Certificate limit described above. Renewals are still subject to the Duplicate Certificate limit.

So (from what I understood) each pre-existing discrete certificate (same set of domains) can be renewed up to five times a week. @cpu did I get this right ?

> Yeah, originally I wanted to cap it to a lower value for the stated reasons, but then comes the problem of testing. How can the test unit test this feature by "simulating" a passing of 48 hours in a reasonable time?

Rigth, I did not think about this. User adjustable `MAX_VALIDITY` perhaps ?

Greek64:
> So (from what I understood) each pre-existing discrete certificate (same set of domains) can be renewed up to five times a week. @cpu did I get this right ?

From what I understood, the first 5 renewals count as Duplicates and the rest as normal renewals. (In the sense that Duplicate certificates can coexist [new key], but renewed certificates not [same key]).
Else this example given further down does not really make sense to me:

>You can issue 50 certificates in week 1, 50 more certificates in week 2, and so on, while not interfering with renewals of existing certificates.

Nevertheless I would be interested in what @cpu has to say.

> Rigth, I did not think about this. User adjustable `MAX_VALIDITY` perhaps ?

Well, the thing is that LE CA always gives certificates with a lifespan of 90 days (not adjustable through simp_le). So even when adjusting `MAX_VALIDITY` we would have to somehow trick simp_le to think that the period given through `--min_valid` has been reached.

Nevertheless I was about to suggest user adjustable `MAX_VALIDITY` for an entirely different reason.
Namely, as it currently stands, we make the assumption that *all* ACME compliant CAs issue certificates with a lifespan of 90 days. The `--force-renew` function is based upon this assumption.  
But as we allow the user to define an ACME CA other than LE through the `ACME_CA_URI` environment variable - and we cannot guarantee what lifetimes these CAs will issue - a user may run into unexpected behavior when doing `--force-renew` on a different CA.  
So my suggestion would be to have a user adjustable `MAX_VALIDITY` variable that *must* be set by the user whenever the `ACME_CA_URI` is set.

Greek64:
Another idea would be to *modify* the `localtime` of the letsencrypt-companion docker container. (e.g. by mounting a custom `localtime` inside the container, that we can modify from the outside)  
This would not affect anything on the host system.

Merry Christmas, btw (Or Happy Holidays)

buchdag:
> Well, the thing is that LE CA always gives certificates with a lifespan of 90 days (not adjustable through simp_le). So even when adjusting `MAX_VALIDITY` we would have to somehow trick simp_le to think that the period given through `--min_valid` has been reached.

You might have mixed up things here, or I did. If the `MAX_VALIDITY` is adjustable, it can default to say 88 days and we can manually set it to 90 days for the test so that the rest of the existing test unit will work as is (and we don't have to fiddle with the container `localtime`).

Maybe `MAX_VALIDITY` should be renamed `MIN_VALIDITY_CAP`, that might be a bit clearer.

Regarding the improved support of other ACME CA I was about to answer that I'm not aware of any other CA offering free certificate through ACME v1 protocol with `http-01` challenge, [but a quick research seems to indicate that there is now at least one candidate](https://www.buypass.com/ssl/products/acme).

Greek64:
> If the `MAX_VALIDITY` is adjustable, it can default to say 88 days and we can manually set it to 90 days for the test so that the rest of the existing test unit will work as is (and we don't have to fiddle with the container `localtime`).

Oh, yes of course. That would work.
But that would mean, that we trust the user to give a value that has already taken account the rate limits (e.g. 88 days instead of 90), *IF* he sets the variable manually and doesn't use the default value.
Either way it's his fault at this point, as long as we document it.

> Regarding the improved support of other ACME CA I was about to answer that I'm not aware of any other CA offering free certificate through ACME v1 protocol with `http-01` challenge, [but a quick research seems to indicate that there is now at least one candidate](https://www.buypass.com/ssl/products/acme).

And according to them they give certificates with a lifetime of 180 days.
So, what do you say? (I could integrate the check in this PR)

buchdag:
> we trust the user to give a value that has already taken account the rate limits (e.g. 88 days instead of 90), IF he sets the variable manually and doesn't use the default value.

Maybe it should be documented that this value is (for now) only adjustable for test purpose.

> And according to them they give certificates with a lifetime of 180 days.
> So, what do you say? (I could integrate the check in this PR)

They also seems to have a limit of one domain per certificate:

> This is a Domain Validated (DV) certificate including a single domain name per certificate.

So we'll probably have to add checks for that to if we intend to support this ACME CA as well. I'll do some tests with Buypass Go as soon as possible and we'll handle the "other ACME CA" compatibility in another, later PR.

app/letsencrypt_service
```diff
@@ -8,6 +8,8 @@ ACME_CA_URI="${ACME_CA_URI:-https://acme-v01.api.letsencrypt.org/directory}"
 DEFAULT_KEY_SIZE=4096
 REUSE_ACCOUNT_KEYS="$(lc ${REUSE_ACCOUNT_KEYS:-true})"
 REUSE_PRIVATE_KEYS="$(lc ${REUSE_PRIVATE_KEYS:-false})"
+MAX_VALIDITY=7776000
```
buchdag: Maybe rename this variable to `MIN_VALIDITY_CAP` (or something else along the same line) to avoid possible confusion with the default duration of certs generated by LE / the ACME CA used ?

README.md
Greek64: done

app/letsencrypt_service
Greek64: done

Greek64:
Did all the mentioned changes.
What's left now is to discuss the aforementioned lower bound

> I also originally wanted to add also a lower bound, to prevent unwanted downtime due to misconfiguration, since the check is triggered hourly and the user could give a period smaller than 3600 seconds.

buchdag:
@Greek64 thanks for the notification. I'm currently bed ridden due to flu, so merge will have to wait a bit. Don't want to merge stuff while my brain is half functional.

Additionally I thought ~ one week ago that maybe the adjustable  `MIN_VALIDITY_CAP` for test purpose wasn't the cleanest idea : [we could just change the test boulder config to issue certificates with a 88 days validity](https://github.com/letsencrypt/boulder/blob/9afa0f7f1967fd4b8c63f4a100a2795461b36609/test/config/ca-a.json#L39). We're already [touching the config](https://github.com/JrCs/docker-letsencrypt-nginx-proxy-companion/blob/6a90d53c975a8d44539d6e9e8737c70150354aca/test/setup/setup-boulder.sh#L13) to change ports for the `http-01` challenge back to the usual 80/443 and whitelist the test domains, so adding another change should be easy.

Your thought on this ?

Greek64:
>I'm currently bed ridden due to flu

I'm sorry to hear that.  
Get well soon. 

>we could just change the test boulder config to issue certificates with a 88 days validity

That seems indeed to be the cleanest way. Didn't even think of touching the Boulder...
So, I will redo this PR in the next few days.

Greek64:
Although, now that I think about it again, we still have a problem.
If we edit the `ca-a.json` and `ca-b.json` the same way we modify the `va.json` for the ports, the change would be made with sed during the test setup phase. 

That means that all tests will get certificates with a validity of 88 days.
And that will conflict with the `force_renew` test, which assumes a 90 day validity in order for it to work.

If we want to make it with the boulder change we either have to add an exception code inside `letsencrypt-service` to use a 88 day validity for the force renew command during a test, or restart the boulder with different settings in between the tests.

cpu:
I'm catching up on email/github from the holiday period and only quickly skimmed this thread. Apologies if I'm misunderstanding something.

> From what I understood, the first 5 renewals count as Duplicates and the rest as normal renewals. (In the sense that Duplicate certificates can coexist [new key], but renewed certificates not [same key]).

The public key of the certificate doesn't have any bearing on whether or not Let's Encrypt considers it a renewal. Certificate B is considered a renewal of Certificate A if and only if the set of subject identifiers (CN and SANs) between A and B are an exact match. If Certificate A and B have the same names but different public keys Certificate B is still a renewal for rate limiting purposes. If Certificate B added or removed a name it would not be considered a renewal. 

> If we want to make it with the boulder change we either have to add an exception code inside letsencrypt-service to use a 88 day validity for the force renew command during a test, or restart the boulder with different settings in between the tests.

I think testing this in an end-to-end integration testing scenario with Boulder will be tricky for the reasons you describe. 

It might be helpful to know that Boulder's existing integration tests take the same approach you describe here (starting Boulder with one set of configs, doing some setup, tearing Boulder down, changing the config, and then restarting Boulder) to simulate the passage of time. See: https://github.com/letsencrypt/boulder/blob/9afa0f7f1967fd4b8c63f4a100a2795461b36609/test/integration-test.py#L722-L737 (`setup_seventy_days_ago` and `setup_twenty_days_ago()` run Boulder with the `FAKECLOCK` configured accordingly).

buchdag:
> And that will conflict with the `force_renew` test, which assumes a 90 day validity in order for it to work.

Hmmm as I view it the test will still work, `/app/force_renew` script only assume that the default cert validity is X (seconds days or whatever) and run `simp_le` on the existing certificates with a minimum required validity greater than or equal to X, hence forcing the renewal.

The minimum validity of the `/app/force_renew` script might as well be 300 days or even 10 years, as long as it's above the default validity of the issued certs (wether 90 days for real LE certs or 88 days for our boulder certs), it would work the same.

We just have to make sure that changes made still allow the `/app/force_renew` to use a `--min-validity` of 90 days or above.

This seems to be the same type of confusion as https://github.com/JrCs/docker-letsencrypt-nginx-proxy-companion/pull/485#issuecomment-449719052. The different time value relationship and current values (in seconds) is as follow (in seconds) :

0 **<**  A (7200) **≤** B (`LETSENCRYPT_MIN_VALIDITY`) **≤** C/D (7603200) **<**  E (7776000) **≤** F (7776000)

- A: lower bound to `LETSENCRYPT_MIN_VALIDITY`
- B: user defined or default `LETSENCRYPT_MIN_VALIDITY` value
- C: upper bound to `LETSENCRYPT_MIN_VALIDITY`
- D: default validity of cert issued by the test suite's Boulder
- E: Let's Encrypt default cert validity
- F: value of the `--min-validity` flag used by `/app/force_renew`

Greek64:
> This seems to be the same type of confusion as [#485 (comment)](https://github.com/JrCs/docker-letsencrypt-nginx-proxy-companion/pull/485#issuecomment-449719052).

Yes indeed. 
Dunno what's the problem with me...
Thanks for the clear up.

Greek64:
Done.
Interestingly some times a wrong certificate is issued during tests and fail, which is *fixed* by a re-run.
Because the first travis run failed, I just force pushed again to re-trigger travis and the checks now pass.

buchdag:
Yes, I'm aware of this, and I haven't managed to understand why this is happening yet.

buchdag:
Oh, one last thing, I've noticed that for some reason your commit messages are prefixed with a star and a space (`* `), could you please remove that ?

Greek64:
>I've noticed that for some reason your commit messages are prefixed with a star and a space (`* `)

That's just the commit message convention that I use.

>I'd prefer that to be removed if possible.

Done

buchdag:
Thanks for the good work 👍

buchdag:
@Greek64 the test unit fails quite often, if you have some spare time could you try to investigate and see if it could be made a bit more reliable ?

Greek64:
Will do. (although until now I had it never fail before... Then again I'm mostly testing with the two container setup...)

Greek64:
@buchdag 
As a quick fix, do you want to try if increasing the sleep time in [Line 63](https://github.com/JrCs/docker-letsencrypt-nginx-proxy-companion/blob/cff8e96da6a1ed4670a364932632e4f1f88edcb3/test/tests/certs_validity/run.sh#L63) to e.g. `sleep 15` increases reliability in your setup?

Greek64:
I ran the `certs_validity` test 10 consecutive times in a three container setup, and it passed every time.

It will be difficult for me to debug this.
Nevertheless I have two possible solutions, although you would have to test them.

First a small Explanation of what the tests does
------------------------------------------------
The test unit creates 3 containers of which only one (`le3.wtf`) has its `LETSENCRYPT_MIN_VALIDITY` set to something less than the lifetime of the issued certificates (specifically 10 seconds less).
After waiting for the symlinks of the certificates and parsing the current lifetime of the issued certificates, the test unit waits 10 seconds, then retriggers `update_certs`, reparses the lifetime of the certificates and checks if the certificate for `le3.wtf` was renewed.

Possible Solution 1
-------------------
Since the `LETSENCRYPT_MIN_VALIDITY` is set to a value 10 seconds less than the lifetime of the issue certificates, if the "waiting for symlinks" and "parsing lifetimes of issued certificates" takes less than a second the `update_certs` script will be triggered too close to the threshold and may not renew the certificate. Thus increasing the `sleep` timeout (like described above) should *fix* that.

Possible Solution 2
-------------------
The certificates are parsed again immediately after the triggering of the `update_certs` script for their lifetimes, which may be too soon (i.e. before the certificate was actually renewed). Putting a `sleep` between the two calls should mitigate this (even better would be a wait function that waits for the cert to be renewed). Though if this is the fix, we should probably also update the `force_renew` test, since it does exactly the same.

buchdag:
I haven't seen CI errors related to this in weeks, so I guess this is no longer relevant. The errors are probably more Travis related than anything.

PR #502, opened 2019-01-22
fix syntax in step 3

mujahed1987:
adding missed " \"  ### Step 3 - proxyed container(s) after ```---name your-proxyed-app```


buchdag:
Merged, thank you !

PR #516, opened 2019-02-28
Correctly retry if format is not expected

### LINKED ISSUE #514: Container ID does not work reliably in fedora atomic systems
cgroup and cpuset in fedora atomic seem to start with `/system.slice/....` breaking the current container id discovery.

The current fallback mechanism only takes the existence of files into account and doesn't handle whether the name has a format which it can handle.

xiamaz:
If the format of a CID source is unable to be matched, the next source should be tried.

Also try to exactly match 64bit alnum strings. Which should work be more robust against smaller format changes.

This should fix #514.


buchdag:
Many thanks, sorry for the merging delay !

Issue #569, opened 2019-08-19
Disable auth_request for location /.well-known/acme-challenge/


dsullivan:
For issues #112 / #232, we ensure that basic authentication is disabled for the challenge location. We should also ensure that auth_request is off.

`location /.well-known/acme-challenge/ {
    auth_request off;
    auth_basic off;
    ...
}`

PR #570, opened 2019-08-19
Set auth_request off for acme challenge location

dsullivan:
Issue #569. Simple change to ensure that auth_request is off for the acme challenge location, just as we do for auth_basic.

### DIFF
--- a/app/nginx_location.conf
+++ b/app/nginx_location.conf
@@ -1,5 +1,6 @@
 location ^~ /.well-known/acme-challenge/ {
     auth_basic off;
+    auth_request off;
     allow all;
     root /usr/share/nginx/html;
     try_files $uri =404;


buchdag:
Seems good to me but I have one question:

[Nginx doc for `auth_request`](https://nginx.org/en/docs/http/ngx_http_auth_request_module.html) says that

> This module is not built by default, it should be enabled with the `--with-http_auth_request_module` configuration parameter.

What are the consequences of including statements for a potentially unbuilt module ?

dsullivan:
Argh.

I had to compile from source to find a nginx install without the auth_request module enabled. (It seems to be enabled for in the populate Docker images, Alpine and Ubuntu packages, etc.) But, yes, without the module enabled, nginx throws an error:

`nginx: [emerg] unknown directive "auth_request" in /usr/local/nginx/conf/nginx.conf:38`

And from what I understand, nginx doesn't have an equivalent of Apache's <IfModule> directive. There's probably no way for this container's scripts to interrogate the nginx container for available modules, either.

Although in practice I don't imagine anyone would run into problems, it's probably not worth the risk.

buchdag:
Is this Nginx doc really up to date ? Because none of the Nginx container Dockerfiles include `--with-http_auth_request_module` yet the `auth_request` module seems to be there.

buchdag:
Ok, bottom line is while Nginx is not built with the `http_auth_request` module by default, most distro packaged versions appears to include it (along with kinda useful stuff that isn't included by default either like the SSL module).

So the containerised versions of Nginx (either alpine or debian, and as far back as the tagged versions go) also include this module :

```term
[buchdag@host ~]$ docker run nginx:1.7 nginx -V
nginx version: nginx/1.7.12
built by gcc 4.7.2 (Debian 4.7.2-5) 
TLS SNI support enabled
configure arguments: --prefix=/etc/nginx --sbin-path=/usr/sbin/nginx --conf-path=/etc/nginx/nginx.conf --error-log-path=/var/log/nginx/error.log --http-log-path=/var/log/nginx/access.log --pid-path=/var/run/nginx.pid --lock-path=/var/run/nginx.lock --http-client-body-temp-path=/var/cache/nginx/client_temp --http-proxy-temp-path=/var/cache/nginx/proxy_temp --http-fastcgi-temp-path=/var/cache/nginx/fastcgi_temp --http-uwsgi-temp-path=/var/cache/nginx/uwsgi_temp --http-scgi-temp-path=/var/cache/nginx/scgi_temp --user=nginx --group=nginx --with-http_ssl_module --with-http_realip_module --with-http_addition_module --with-http_sub_module --with-http_dav_module --with-http_flv_module --with-http_mp4_module --with-http_gunzip_module --with-http_gzip_static_module --with-http_random_index_module --with-http_secure_link_module --with-http_stub_status_module --with-http_auth_request_module --with-mail --with-mail_ssl_module --with-file-aio --with-http_spdy_module --with-cc-opt='-g -O2 -fstack-protector --param=ssp-buffer-size=4 -Wformat -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2' --with-ld-opt='-Wl,-z,relro -Wl,--as-needed' --with-ipv6
[buchdag@host ~]$ docker run nginx:1.8-alpine nginx -V
nginx version: nginx/1.8.1
built by gcc 5.3.0 (Alpine 5.3.0) 
built with OpenSSL 1.0.2g  1 Mar 2016
TLS SNI support enabled
configure arguments: --prefix=/etc/nginx --sbin-path=/usr/sbin/nginx --conf-path=/etc/nginx/nginx.conf --error-log-path=/var/log/nginx/error.log --http-log-path=/var/log/nginx/access.log --pid-path=/var/run/nginx.pid --lock-path=/var/run/nginx.lock --http-client-body-temp-path=/var/cache/nginx/client_temp --http-proxy-temp-path=/var/cache/nginx/proxy_temp --http-fastcgi-temp-path=/var/cache/nginx/fastcgi_temp --http-uwsgi-temp-path=/var/cache/nginx/uwsgi_temp --http-scgi-temp-path=/var/cache/nginx/scgi_temp --user=nginx --group=nginx --with-http_ssl_module --with-http_realip_module --with-http_addition_module --with-http_sub_module --with-http_dav_module --with-http_flv_module --with-http_mp4_module --with-http_gunzip_module --with-http_gzip_static_module --with-http_random_index_module --with-http_secure_link_module --with-http_stub_status_module --with-http_auth_request_module --with-mail --with-mail_ssl_module --with-file-aio --with-http_spdy_module --with-ipv6
```

I don't see why the `http_auth_request` module would disappear from either Alpine's or Debian's Nginx anytime soon, so I guess this can be safely merged after all.

What is your view on this ?

dsullivan:
I also was unable to find any packaged version of nginx that did not include the http_auth_request module. I had to compile from source myself to test the error condition.

I suspect that anyone who compiles nginx themselves probably has their own certificate management strategy, too. So I think this is safe to merge.

buchdag:
Agreed, merging. Thanks for the contribution.

Issue #561, opened 2019-07-28
config not generated for asterisk domain


sandrodz:
<img width="524" alt="Screen Shot 2019-07-27 at 8 09 59 PM" src="https://user-images.githubusercontent.com/8479569/62003573-fee97900-b129-11e9-82ea-9cc2fec7f4eb.png">

buchdag:
Hi, thanks for reporting this, I'm working on it.

buchdag:
Can you pull and try `duch/letsencrypt-nginx-proxy-companion:561` to see if the fix work for you ?

If yes I'll make a PR and merge this branch.

I made the location configuration work as follow. Let say the domain is `foo.bar.baz.example.tld`:

- if a config file with the exact same name as your domain (`foo.bar.baz.example.tld`) is present, use it.
- else try every corresponding wildcard domain, from most qualified to least qualified, first ascending through the labels to the base domain (`*.bar.baz.example.tld`, `*.baz.example.tld`, `*.example.tld`) then descending (`foo.bar.baz.example.*`, `foor.bar.baz.*`, `foo.bar.*`, `foo.*`). The first that match an existing config file will be used.
- if no wildcard location config file was found either, use the `default` location config file.

buchdag:
ping @sandrodz

sandrodz:
Hey, I'm at loss, how would I test this? I cannot find this image: `duch/letsencrypt-nginx-proxy-companion:561`

buchdag:
Sorry, last week push seem to have failed without me noticing it.

I pushed the image again, it's now available : https://cloud.docker.com/repository/docker/duch/letsencrypt-nginx-proxy-companion/tags

smartens:
I also need some guidance for this. I trying to get gitlab pages working. I tried your image but it still doesn't work. Here is my docker compose file:

If you need any logs, please let me know.

```
version: '3'

networks:
  nginx_default:
    external: true
    
volumes:
  config:
  logs:
  data:

services:
  gitlab:
    image: gitlab/gitlab-ce:latest
    container_name: gitlab
    restart: unless-stopped
    #hostname: git.team.softec.uni-due.de
    networks:
      - nginx_default
    environment:
      VIRTUAL_HOST: git.team.softec.uni-due.de,*.pages.team.softec.uni-due.de
      LETSENCRYPT_HOST: git.team.softec.uni-due.de,*.pages.team.softec.uni-due.de
[...]

PR #588, opened 2019-10-10
Enable use of wildcard location configurations

buchdag:
This PR is a fix for #561 

It makes the location configuration work as follow. Let say the domain is `foo.bar.baz.example.tld`:

- if a config file with the exact same name as your domain (`foo.bar.baz.example.tld`) is present, use it.
- else try every corresponding wildcard domain, from most qualified to least qualified, first ascending through the labels to the base domain (`*.bar.baz.example.tld`, `*.baz.example.tld`, `*.example.tld`) then descending (`foo.bar.baz.example.*`, `foor.bar.baz.*`, `foo.bar.*`, `foo.*`). The first matching config file will be used.
- if no wildcard location config file was found either, use the `default` location config file.

### DIFF
--- a/app/functions.sh
+++ b/app/functions.sh
@@ -23,10 +23,61 @@ function check_nginx_proxy_container_run {
 fi
 }
 
+function ascending_wildcard_locations {
+    # Given foo.bar.baz.example.com as argument, will output:
+    # - *.bar.baz.example.com
+    # - *.baz.example.com
+    # - *.example.com
+    local domain="${1:?}"
+    local first_label
+    regex="^[[:alnum:]_\-]+(\.[[:alpha:]]+)?$"
+    until [[ "$domain" =~ $regex ]]; do
+      first_label="${domain%%.*}"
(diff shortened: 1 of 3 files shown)


buchdag:
+ corresponding tests

jayhar:
Previously to achieve a wildcard SSL for all subdomains I have been using the following example. Two A records are needed to make this work. Note I do not force redirect from non-www to www. 

docker run -d -e VIRTUAL_HOST=.example.com -e LETSENCRYPT_HOST=www.example.com,example.com -e LETSENCRYPT_EMAIL=example@example.com --network=exampleproxy --name example.com -v /srv/www/nginx-data/html/example.com:/usr/local/apache2/htdocs httpd:alpine

With the changes you made (thank you for your work) am I am able to to change my current syntax to -e VIRTUAL_HOST=example.com -e LETSENCRYPT_HOST=example.com to achieve the same scenario? The issue I had before is the NGINX config had to manually modified to make this work.

PR #601, opened 2019-12-09
Create Deploy-in-Swarm.md

Maxthod:
This is a docker-compose that allow to deploy in the swarm very easily. It used the image "helder/docker-gen", which I've found in your doc.
Works great!


buchdag:
Hi. We don't directly provide compose file examples in the docs. Feel free to host your own compose file example where you see fit and submit a documentation PR with a link to it.

Issue #410, opened 2018-06-28
How to explicitly not create a SAN certificate?


erikverheij:
Is it possible to instruct the companion to create one certificate per domain? In our setup it's preferable as we don't want www.example.org to be dependent on the availability of example.com.

```
- VIRTUAL_HOST=www.example.org,example.org,www.example.com,example.com
- LETSENCRYPT_HOST=www.example.org,example.org,www.example.com,example.com
```

buchdag:
Unfortunately no, that's not possible at the moment, but that might be a use case for [the experimental standalone cert feature](https://github.com/JrCs/docker-letsencrypt-nginx-proxy-companion/wiki/Standalone-certificates-(beta)), if you are willing to give it a try.

mastef:
Just came here looking for this 👍  Would be great, so we wouldn't have to have 100 containers with one customer cert each, but one container that can handle 100 domains that have different certs. ( We have a client-side SPA - every client is being served the same read only html content, just on their own domain )

buchdag:
@ivolution643 thanks for spotting that, I've just pushed a version of `dev` [that should take care of it](https://github.com/JrCs/docker-letsencrypt-nginx-proxy-companion/blob/176d98fc5e9ce160a836f1efdb7ebd612e92c49c/app/functions.sh#L41) (it does according to the short test I did).

Don't hesitate to try again with the fixed version and to report back (just wait for the new `dev` tag autobuild to complete on Dockerhub).

buchdag:
Feature added by @SilverFire in #647

Issue #368, opened 2018-04-24
Standalone Certificate


flocomkoko:
Is it possible to request a standalone certificate for further use (in my case a simple Mail server), that is autorenewed by the nginx-letsencrypt container without spinning up an container, that is basically useless?
I was thinking of something like 
`docker exec nginx-letsencrypt /app/request....`
Parallel to this "standalone cert" i would like to use the complete functionality of this nice container including the proxy for other services (in my case for a simple wiki and a nextcloud instance).

cphamlet:
I would also like to know this.

cphamlet:
This docker container makes use of the simp_le client, a lightweight letsencrypt client. [https://github.com/zenhack/simp_le](https://github.com/zenhack/simp_le )

You can try running that command on start, and stick it in the auto renew code.

buchdag:
An **experimental** version of this feature inspired by @ymettier PR #220 has been added to `dev`.

[This instructions are here](https://github.com/JrCs/docker-letsencrypt-nginx-proxy-companion/wiki/Standalone-certificates-(beta)). Any feedback is welcome.

ladrua:
Just tested this feature, works like a charm. Personally really need this in production yesterday... 😺 
How are the changes to the letsencrypt_user_data file picked up/triggered though?

buchdag:
@ladrua for now they're not picked up at all, meaning you'll either have to

- wait for the next service loop execution (every hour)
- manually trigger it with `docker exec your-le-container /app/signal_le_service`

ladrua:
Thanks. Thats fine. I had some issues all of a sudden on a new build of dev, had to `RUN chmod +x /app/signal_le_service && chmod +x /app/letsencrypt_service` because I got Permission Error. Also after, the chalange configs are left behind in the conf.d(the `example.com-standalone.conf` folder, and we get errors due to double definition of hostnames(we have other configs for the actuall services).

buchdag:
@ladrua I pushed two more commits to `dev`. The missing `;` has been corrected and the challenge config should be correctly removed too. No idea where you permission issue came from, I'll have to take a look again.

ladrua:
@buchdag Thanks! I'll have a go later. Also, does it create the challenge configs even though renewal is not necessary? Must it? Just realized that this recreation happend every time(hour) the service was triggered. So if I manually deleted the standalone confs, they came back an hour later.

ladrua:
[...]

PR #610, opened 2020-01-09
Implemented LETSENCRYPT_STANDALONE_CERTS support from container env

SilverFire:
Hi, @buchdag 

I've followed your idea with LETSENCRYPT_STANDALONE_CERTS and ported it to environment variables of Nginx container.

The usage will be pretty simple:

docker-compose.yml

```
version: '3'

services:
  nginx:
    image: nginx:latest
    environment:
      - "LETSENCRYPT_STANDALONE_CERTS=true"
      - "VIRTUAL_HOST=example.com,private.acme.com,dev.acme.com"
      - "LETSENCRYPT_HOST=example.com,private.acme.com,dev.acme.com"
```

This config will produce three separate certificates. It seems more correct in terms of decoupling to claim separate certificates from the Nginx container env, than touching the main LE companion.

If the PR is acceptable, I will adjust the documentations

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ nginx.tmpl
 test/local_test_env.sh
 test/tests/docker_api/expected-std-out.txt
 test/tests/container_restart/docker_event_out.txt
+test/tests/certs_standalone/letsencrypt_user_data
--- a/app/entrypoint.sh
+++ b/app/entrypoint.sh
@@ -160,6 +160,7 @@ if [[ "$*" == "/bin/bash /app/start.sh" ]]; then
     check_writable_directory '/etc/nginx/certs'
     check_writable_directory '/etc/nginx/vhost.d'
     check_writable_directory '/usr/share/nginx/html'
+    [[ -f /app/letsencrypt_user_data ]] && check_writable_directory '/etc/nginx/conf.d'
     check_deprecated_env_var
     check_default_cert_key
     check_dh_group
--- a/app/functions.sh
+++ b/app/functions.sh
@@ -96,6 +96,39 @@ function add_location_configuration {
     fi
 }
 
+function add_standalone_configuration {
+    local domain="${1:?}"
+    if grep -q $domain "/etc/nginx/conf.d/default.conf"; then
+        # If the domain is already present in nginx's conf, use the location configuration.
+        add_location_configuration "$domain"
+    else
+        # Else use the standalone configuration.
+        cat > "/etc/nginx/conf.d/standalone-cert-$domain.conf" << EOF
+server {
+    server_name $domain;
+    listen 80;
+    access_log /var/log/nginx/access.log vhost;
+    location ^~ /.well-known/acme-challenge/ {
+        auth_basic off;
+        allow all;
+        root /usr/share/nginx/html;
+        try_files \$uri =404;
+        break;
+    }
+}
+EOF
+    fi
+}
+
+function remove_all_standalone_configurations {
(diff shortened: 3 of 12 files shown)


buchdag:
Hi.

Sorry if I'm not getting it but how is that any different from using a dummy container to generate certificates ?

buchdag:
The idea behind the standalone cert feature was to generate certificate that aren't tied to any specific container's Environment.

SilverFire:
I didn't see the dummy containers idea in the documentation. Do you have such?

There are a few use cases I address with this PR:

1. Container that listens to 10-20 domain names, but I don't want to have a common certificate for them since they represent different white labels.
2. I don't want to reconfigure LE companion when I change the domains list
3. If one of the domains fails to obtain certificate because DNS is messed up, I don't want other certificates to fails.

buchdag:
Oh, it sounds like an attempt to implement something similar to https://github.com/JrCs/docker-letsencrypt-nginx-proxy-companion/issues/410 then.

Did I get that right ?

SilverFire:
Yes, it actually implements issue #410, but with a little bit simpler way of configuration.

Instead of having something like this 

```
LETSENCRYPT_STANDALONE_CERTS=('foo_com' 'bar_com', 'www_bar_com')
LETSENCRYPT_foo_com_HOST=foo.com
LETSENCRYPT_foo_com_EMAIL=letsencrypt@foo.com
LETSENCRYPT_bar_com_HOST=bar.com
LETSENCRYPT_bar_com_EMAIL=letsencrypt@bar.com
LETSENCRYPT_www_bar_com_HOST=www.bar.com
LETSENCRYPT_www_bar_com_EMAIL=letsencrypt@bar.com
```

I suggest to have

```
LETSENCRYPT_STANDALONE_CERTS=true
LETSENCRYPT_HOST=foo.com,bar.com,www.bar.com
LETSENCRYPT_EMAIL=letsencrypt@foo.com
```

The only drawback I see is that we can not merge into a single SAN group of domains _foo.com / www.foo.com_, but we can enhance it later upon request.

buchdag:
Ok, standalone certs (#368) and the splitting of SAN certificates (#410) are two completely different features, hence the confusion.

Standalone certs refers to the ability to issue certificate that aren't tied to a specific container's environment.

You PR will have to use another env var than `LETSENCRYPT_STANDALONE_CERTS` as that's reserved for the standalone certs feature and a different term than 'standalone'.

`LETSENCRYPT_SPLIT_CERTS` ?
`LETSENCRYPT_SINGLE_CERTS` ?

SilverFire:
I'd vote for `LETSENCRYPT_SINGLE_CERTS` or a bit more specific: `LETSENCRYPT_SINGLE_DOMAIN_CERTS`

buchdag:
I'm okay with `LETSENCRYPT_SINGLE_DOMAIN_CERTS`.

Don't forget to create the documentation for the now environment variable.

buchdag:
Other than that, the PR look fine to me.

SilverFire:
Wilco

buchdag:
Didn't see you based you branch on `dev`, looks like I fucked up things a bit when I force pushed an up to date version of `dev` thee days ago, sorry for that 😒

buchdag:
@SilverFire could you re-open this PR with a branch that isn't based on `dev` ?

I'd like to merge the standalone feature of `dev` soonish and this is blocking me, plus the previously mentioned issue.

SilverFire:
Will do it today eve

SilverFire:
Sorry for the delay. Replaced with #647

PR #657, opened 2020-05-15
s/proxyed/proxied/g

noahbliss:
;)


buchdag:
Woops, I think "proxyed" is all over the place in almost all the docs. Thanks for the correction.

PR #673, opened 2020-06-30
feat: docker_api https fix for secured docker

alin-o:
The certificate path is also needed when docker daemon mode is `tlsverify`

see https://docs.docker.com/engine/security/https/#connecting-to-the-secure-docker-port-using-curl


buchdag:
Hi.

Thanks for the contribution. Have you tested it ?

Do you think you could add documentation for this feature to your PR ?

alin-o:
Hi,

I tested this on my own servers, which are configured with daemon mode `tlsverify`, and it works. Before this change, no certificates were ever generated because the `curl` command returned just an error message,

I added a short documentation under `docs/Container-configuration.md`. I hope it is the right place.

buchdag:
> I added a short documentation under `docs/Container-configuration.md`. I hope it is the right place.

Yep, that's perfect, thanks.

> I tested this on my own servers, which are configured with daemon mode `tlsverify`, and it works.

Okay this is going to be pretty hard to test on Travis.

buchdag (changes requested):
Somehow I wrote change requests for this PR almost 4 years ago and never pressed the **Submit review** button. 🤦

buchdag:
Implemented in #1265

PR #711, opened 2020-11-12
Search in all config files for standalone config

remogloor:
Search in all .conf files for the server configuration when a standalone configuration exists and not just in default.conf,


buchdag:
LGTM, thanks.

PR #712, opened 2020-11-12
Cleanup location changes

remogloor:
Location changes are not propely undone at the moment.
This is a problem in case you forward requests to an upstream reverse proxy and that one once to use http challenges to create certificates too.


buchdag:
@remogloor ping on the review

buchdag:
No updates from PR creator, closing.

PR #717, opened 2020-11-23
Speed up tests execution.

buchdag:
This PR does two things motsly aimed at speeding up tests:

- It makes `docker-gen`'s `--wait` argument user configurable via an environment variable (`DOCKER_GEN_WAIT`) and lower the default from a **minimum wait of 15 seconds** after container change events to a more reasonable 500 milliseconds. `--wait` was only supposed to act as debounce, and 15 seconds is pretty long for a debounce.
- It uses more agressive polling in the tests and fix the timeouts so that they're based on real elapsed time and not on the number of iterations through the loop.

This saves roughly another 20% of build time on Travis CI.

### DIFF
--- a/app/start.sh
+++ b/app/start.sh
@@ -18,7 +18,9 @@ trap 'term_handler' INT QUIT TERM
 /app/letsencrypt_service &
 letsencrypt_service_pid=$!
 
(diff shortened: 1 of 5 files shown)


buchdag:
Travis CI build cancelled, it already passed on my fork : https://travis-ci.com/github/buchdag/docker-letsencrypt-nginx-proxy-companion/builds/203458199

PR #724, opened 2020-12-07
Allow ZEROSSL_API_KEY on the letsencrypt container

bgarret:
I was trying the ZeroSSL support and the API key wasn't picked up properly by the letsencrypt container, which had me scratching my head.

I found out that setting `ZEROSSL_API_KEY` directly on the letsencrypt container wasn't supported (the docs aren't clear about this), this pull request tries to fix this.

By the way, thanks for all the work, it is much appreciated!


buchdag:
LGTM, thanks for the contribution. 👍 

I agree that the doc could probably be clearer, I'll try to fix this soonish but any improvement suggestions are welcome 😃

PR #875, opened 2021-09-26
Add `DEFAULT_KEY_SIZE` variable to the documentation

HPPinata:
Useful but undocumented environment variable to avoid setting `LETSENCRYPT_KEYSIZE` on every container manually


buchdag:
Thanks ! 👍

PR #882, opened 2021-10-13
Make dhparam-generation consistent with nginx-proxy

Momi-V:
Update keysize, speedup dhparam-generation and make generation consistent with changes made in this commit:
https://github.com/nginx-proxy/nginx-proxy/pull/1452/files/be107963a7dc931322ba90f7d40d5feb46dd13b2

### DIFF
--- a/app/entrypoint.sh
+++ b/app/entrypoint.sh
@@ -47,7 +47,7 @@ function check_writable_directory {
 function check_dh_group {
     # Credits to Steve Kamerman for the background Diffie-Hellman creation logic.
     # https://github.com/nginx-proxy/nginx-proxy/pull/589
-    local DHPARAM_BITS="${DHPARAM_BITS:-2048}"
+    local DHPARAM_BITS="${DHPARAM_BITS:-4096}"
     re='^[0-9]*$'
     if ! [[ "$DHPARAM_BITS" =~ $re ]] ; then
        echo "Error: invalid Diffie-Hellman size of $DHPARAM_BITS !" >&2
@@ -87,7 +87,7 @@ function check_dh_group {
     # Generate a new dhparam in the background in a low priority and reload nginx when finished (grep removes the progress indicator).
     (
         (
-            nice -n +5 openssl dhparam -out "${DHPARAM_FILE}.new" "$DHPARAM_BITS" 2>&1 \
+            nice -n +5 openssl dhparam -dsaparam -out "${DHPARAM_FILE}.new" "$DHPARAM_BITS" 2>&1 \
             && mv "${DHPARAM_FILE}.new" "$DHPARAM_FILE" \
             && echo "Info: Diffie-Hellman group creation complete, reloading nginx." \
             && set_ownership_and_permissions "$DHPARAM_FILE" \


HPPinata:
Closed due to the great work in this pull request
https://github.com/nginx-proxy/nginx-proxy/pull/1797

polarathene:
I will not have time to bring the changes over to this project from what I can tell as I'm rather swamped elsewhere.

Technically there shouldn't be any need for the DH support here AFAIK, unless you need to frequently configure different DH params for each site. With the standard RFC7919 DH groups and the 3 that `nginx-proxy` will now support, I can't think of much reason to require that, other than maybe 1024-bit DH params for legacy, in which case those should be generated.

It should be perfectly fine using the same global DH params for RFC7919 on all sites where possible. 2048-bit is still plenty secure, but feel free to add a bit of overhead with 3072-bit or 4096-bit, most connections should be using ECDHE these days anyway where the DH params aren't used (standardized ECC curves are instead).

HPPinata:
> No problem, I'll handle this 👍

Many thanks!
Do you have any rough estimation on the timeframe until being able to implement those changes?
I am currently in the process of setting up another service and wondering whether it's worth pushing final deployment out a little or if this will take significantly longer (i.e. more than 1-2 months).
Greatly appreciate your work (especially as I am not experienced enough to create anything like this myself).

polarathene:
@HPPinata you probably don't need to worry about DH params, it's only relevant for TLS 1.2, and only for cipher suites that are not ECDHE (which could be used instead). Your client software on your end (or users) most likely supports TLS 1.3, which has it's own DH params built-in, among other cipher suite support.

`acme-companion` still produces DH parameters too, the only change being proposed here is to carry over the change to standard RFC7919 DH groups.

The following line could be changed to match `nginx-proxy` value: `DHPARAM_FILE='/etc/nginx/dhparam/dhparam.pem'`

https://github.com/nginx-proxy/acme-companion/blob/1785bc5394d762eee9492034adfc0c9f8edb6326/app/entrypoint.sh#L59

And that should then be detected as available via `nginx-proxy` already skipping generation. You could alternatively provide the same DH params file at the location that is checked above for the same effect.

All that's going on in that regard is `acme-companion` makes a copy of the DH params file above with the FQDN for `nginx-proxy` to recognize as DH params to use for that container/site instead of the default ones (_even though they might be the same now, so a bit redundant, but it's just a symlink_):

https://github.com/nginx-proxy/acme-companion/blob/1785bc5394d762eee9492034adfc0c9f8edb6326/app/letsencrypt_service#L41-L44

If you want to keep it simple, you should be able to just create a Dockerfile that extends `acme-companion` by adding [any of these DH param files](https://github.com/nginx-proxy/nginx-proxy/tree/main/dhparam) to `/etc/nginx/certs/dhparam.pem`, otherwise volume mount the file locally to that location.

---

> @polarathene the DH support is present in the companion too because there is currently no logic to check if the nginx-proxy container generated a safe DH parameter already.

I guess an easy way to do that if not shipping your own copy of the DH param files is to use a checksum to verify the DH params have the trusted contents that newer versions `nginx-proxy` provide?

Or just [split out the DH handling from `nginx-proxy`](https://github.com/nginx-proxy/nginx-proxy/blob/83e3b25c46c2c99513103a4929e9935f9de7ae24/docker-entrypoint.sh#L65-L95) into it's own script again and have `acme-companion` use `ADD` or similar to pull that script and the DH params over into the Dockerfile build for `acme-companion` to use too?

HPPinata:
I am currently using a Dockerfile to add a secure dhparam.pem file for the acme companion to use (sadly I still need TLS 1.2), but (like the other solutions, especially changes to the source) it adds clutter compared to just pulling an image.
As I still have time I don't want to deploy something with a workaround that will most likely be made redundant in the foreseeable future, therefore the original question about a timeframe at buchdag.

buchdag:
@HPPinata I'm currently blocked on merging new stuff to this project by #887 

I can work on bringing the dh param generation in line with nginx-proxy if you're willing to test the image.

buchdag:
I'm still working on #887

buchdag:
#887 being solved, I'll resume working on this.

Expect it to be done no earlier than next week and no later than the end of the month.

buchdag:
@HPPinata done in #904

PR #904, opened 2021-12-14
Use RFC 7919 DH groups + Remove DH generation

buchdag:
This PR brings acme-companion's DH group creation in line with nginx-proxy : https://github.com/nginx-proxy/nginx-proxy/pull/1797

> This PR replaces the Diffie-Hellman parameter generation with standardized RFC 7919 DH groups (_2048, 3072, 4096 bit lengths_). These files are available from [this repo](https://github.com/internetstandards/dhe_groups) These DH groups are advised to use vs generating your own. The [openssl wiki advises adopting RFC 7919 too](https://wiki.openssl.org/index.php/Diffie-Hellman_parameters#RFC_7919_Groups) and provides the ASCII PEM encoded content with which you can verify against the param files in this PR as the same.

Thanks to @polarathene for the original idea and work.


app/entrypoint.sh
```diff
@@ -45,56 +45,63 @@ function check_writable_directory {
...
+        local USER_PROVIDED_DH
+
+        # Check if the DH params file is user provided or comes from acme-companion
+        local DHPARAM_HASH; DHPARAM_HASH=$(sha256sum "$DHPARAM_FILE" | cut -d ' ' -f1)
+        
+        for f in /app/dhparam/ffdhe*.pem; do
+            local FFDHE_HASH; FFDHE_HASH=$(sha256sum "$f" | cut -d ' ' -f1)
+            if [[ "$DHPARAM_HASH" == "$FFDHE_HASH" ]]; then
+                # This is an acme-companion created DH params file
+                local USER_PROVIDED_DH='false'
```
polarathene, line 81: You already defined and `local` scoped this var earlier in the function, you shouldn't need to use the `local` keyword here again AFAIK?

app/entrypoint.sh
```diff
@@ -45,56 +45,63 @@ function check_writable_directory {
...
+	# DH params may be provided by the user (rarely necessary)
+	if [[ -f ${DHPARAM_FILE} ]]; then
+        local USER_PROVIDED_DH
+
+        # Check if the DH params file is user provided or comes from acme-companion
+        local DHPARAM_HASH; DHPARAM_HASH=$(sha256sum "$DHPARAM_FILE" | cut -d ' ' -f1)
+        
+        for f in /app/dhparam/ffdhe*.pem; do
+            local FFDHE_HASH; FFDHE_HASH=$(sha256sum "$f" | cut -d ' ' -f1)
+            if [[ "$DHPARAM_HASH" == "$FFDHE_HASH" ]]; then
```
polarathene, line 79: There's some inconsistency throughout the diff with vars.

General advice you'd get from a linter like shellcheck is to prefer `"${VAR_NAME}"`, and within `[[ ]]` conditionals you can avoid the quotes IIRC (but not within the other kind `[ ]` that I generally avoid myself).

polarathene (approved):
LGTM besides a few nit-picks.

Also it might not have been visible in your editor due to your tab-width but Github renders a wider tab-width where it's rather obvious that there is mixed indentation of some lines with spaces and others with tabs, you probably want to normalize that.

app/entrypoint.sh
buchdag: I'm already using shellcheck 🙂 

shellcheck

As you can see it's not complaining about `"${VAR_NAME}"` vs `"$VAR_NAME"` and require quotes on the right-hand side of `==` inside `[[ ]]`. Quoting the left-hand side is just an habit I picked up from the defensive / safe / whatever bash programming tenet of "quote all uses of variables".

I'm personally not a big fan of using `"${VAR_NAME}"` over `"$VAR_NAME"`, my personal rule of thumb is to use `{}` only when I'm doing variable manipulation / expansion / interpolation, like [described here](https://github.com/anordal/shellharden/blob/master/how_to_do_things_safely_in_bash.md#should-i-use-curly-braces).

I'll make at least the `check_dh_group` function internally consistant but clearly there are inconsistencies here and there on the whole code base. 😦

app/entrypoint.sh
buchdag: Typo, thanks for spotting this 👍

buchdag:
> Also it might not have been visible in your editor due to your tab-width but Github renders a wider tab-width where it's rather obvious that there is mixed indentation of some lines with spaces and others with tabs, you probably want to normalize that.

No, I'm aware of this, I think it's been present in the code since before I started contributing to this project and I never got to normalizing the situation.

polarathene (approved):

PR #923, opened 2022-02-17
Added DEFAULT_EMAIL env variable to docker-compose docs

marc-hg:
I noticed the readme.md includes an env variable for the letsencrypt email so you can be notified if your certs are about to expire(they should automatically renew with this anyway but still) but the docker-compose doesn't so I thought it might be worth to add it there too

image
image


buchdag:
Thanks !

PR #946, opened 2022-04-11
docs: moved DEFAULT_EMAIL env to acme-companion

RodrigoTomeES:
https://github.com/nginx-proxy/acme-companion#step-2---acme-companion


buchdag:
Thanks !

PR #960, opened 2022-07-11
added docker compose 3 example config for two containers

seekermarcel:
in docker compose version 3 the volumes_from function is removed.
Therefore another way is needed to connect both containers together. One option is to use the NGINX_PROXY_CONTAINER env variable


Pubec (approved):

skamensky:
Thank you for this @seekermarcel !

PR #976, opened 2022-11-18
check EAB_KID and HMAC_KEY before registering the ACME account

a1ad:
A possible fix for https://github.com/nginx-proxy/acme-companion/issues/975
It is working for my use case, please check I didn't break something else.


buchdag:
@a1ad thanks for this fix !

PR #1020, opened 2023-04-02
Added Troubleshooting solution for deleted certs

mkaufhol:
Added a solution to a problem that I faced when I tried to set everything up from scratch after I've got everything running for a few months. In my situation, these few lines could've saved me a lot of time.


buchdag:
Hi. Could you give more details about what you did, what was your configuration, what where the errors messages, how could it be reproduced ? I routinely restart existing setups from scratch and/or delete certificates for testing reasons and I never encountered this issue or had to revoke certificates.

PR #1032, opened 2023-04-30
ci: rename build / publish workflow

buchdag:
We're not publishing on DockerHub only, this PR reflects that.


buchdag:
Already done.

PR #1064, opened 2023-11-23
fix: prevent infinite loop in enumerate_wildcard_locations

egormkn:
This pull request fixes two issues in `enumerate_wildcard_locations` function:

1) When a wildcard domain passed to the function, it never returns. Even though wildcard domains are not supported at the moment, I think, infinite loop should be fixed. Currently any wildcard domain in any container configuration completely hangs the service. 
2) In `descending_wildcard_locations` the first occurrence of `last_label` was removed, which might generate wrong wildcards:
```bash
#!/bin/bash

source ./functions.sh
enumerate_wildcard_locations test.com.example.com

# Output:
# *.com.example.com
# *.example.com
# test.example.com.*  <-- should be test.com.example.*
# test.example.*      <-- should be test.com.*
# test.*
```

### DIFF
--- a/app/functions.sh
+++ b/app/functions.sh
@@ -62,11 +62,12 @@ function ascending_wildcard_locations {
     # - *.example.com
     local domain="${1:?}"
     local first_label
-    regex="^[[:alnum:]_\-]+(\.[[:alpha:]]+)?$"
-    until [[ "$domain" =~ $regex ]]; do
+    tld_regex="^[[:alpha:]]+$"
+    regex="^[^.]+\..+$"
+    while [[ "$domain" =~ $regex ]]; do
       first_label="${domain%%.*}"
-      domain="${domain/${first_label}./}"
-      if [[ -z "$domain" ]]; then
+      domain="${domain/#"${first_label}."/}"
+      if [[ "$domain" == "*" || "$domain" =~ $tld_regex ]]; then
         return
       else
         echo "*.${domain}"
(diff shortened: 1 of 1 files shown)


buchdag:
@egormkn thanks for this fix 👍 

Do you think it would be possible to update the tests so that the bug you're fixing is covered ?

buchdag:
```diff
--- a/test/tests/location_config/expected-std-out.txt
+++ b/test/tests/location_config/expected-std-out.txt
@@ -1,7 +1,9 @@
-*.bar.baz.example.com
-*.baz.example.com
+*.bar.baz.com.example.com
+*.baz.com.example.com
+*.com.example.com
 *.example.com
-foo.bar.baz.example.*
+foo.bar.baz.com.example.*
+foo.bar.baz.com.*
 foo.bar.baz.*
 foo.bar.*
 foo.*
  ```
 ```diff
--- a/test/tests/location_config/run.sh
+++ b/test/tests/location_config/run.sh
@@ -54,7 +54,7 @@ function check_location {
 }
 
 # check the wildcard location enumeration function
-docker exec "$le_container_name" bash -c 'source /app/functions.sh; enumerate_wildcard_locations foo.bar.baz.example.com'
+docker exec "$le_container_name" bash -c 'source /app/functions.sh; enumerate_wildcard_locations foo.bar.baz.com.example.com'
 
 # default configuration file should be empty
 config_path="$vhost_path/default"
```

buchdag (approved):

PR #1074, opened 2023-12-15
fix: dot not remove symlinked default certificate / private key

buchdag:
If `default.crt` / `default.key` were symlinks to certificates managed by this container, they were removed along other symlinks in `/etc/nginx/certs`. This was an unintentional behaviour.


buchdag:
@Exagone313

PR #1078, opened 2023-12-19
fix: handle most recently created containers first

### LINKED ISSUE #1049: Slow creation for new domain
https://github.com/nginx-proxy/acme-companion/blob/9fbd6684efba2d740f195f12888362307d26bb59/app/letsencrypt_service#L432C5-L432C5

This loop can take a long moment if you have lots of domains/containers. This is a NOT a problem for certificate renewal, but it is for certificate CREATION.

This list is ordered (by go) in alphabetical order. If you new domain starts with the letter `V` for instance, you have to wait for ACME script to walk through all previous domains (even if no action required yet) before reaching you newly domain.

This can prevent your domain to be ready for several minutes if you have lots of domains and this time vary a lot based on your domain name (alphabetical order).

- What solution is possible?
- Can this loop be split into multiple chunks and run in parallel? And then using a `wait` before going on?
- Did I reach a point where `nginx + acme-companion` is not a viable solution anymore? What do you recommend if this is the case?

buchdag:
This is a proposed fix for #1049. It orders the containers being iterated over by descending creation timestamp (most recent first) instead of by ascending certificate domain name.

Closes #1049

### DIFF
--- a/app/letsencrypt_service_data.tmpl
+++ b/app/letsencrypt_service_data.tmpl
@@ -1,19 +1,21 @@
+#!/bin/bash
+# shellcheck disable=SC2034
 LETSENCRYPT_CONTAINERS=(
-    {{ $orderedContainers := sortObjectsByKeysDesc $ "Created" }}
-    {{ range $_, $container := whereExist $orderedContainers "Env.LETSENCRYPT_HOST" }}
-        {{ if trim $container.Env.LETSENCRYPT_HOST }}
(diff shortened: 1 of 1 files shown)


buchdag:
@cpontvieux-systra I know this is ~ 4 months old but your help in validating that this fix #1049 would be welcome.

I've pushed and image that include the fix : `nginxproxy/acme-companion:1049`

PR #1063, opened 2023-10-22
Upgrade acme.sh to version 3.0.7

pini-gh:
to benefit from last improvements of script dnsapi/dns_gandi_livedns.sh.

PR #1087, opened 2024-01-14
build: acme.sh 2.9.0 -> 3.0.4

### LINKED ISSUE #1035: [CVE] (possible?) RCE because of acme.sh - update from 2.9.0 to latest version
Bug description
-----------------

This image/ project is based on [acmesh-official/acme.sh](https://github.com/acmesh-official/acme.sh) which had a [CVE with possible RCE](https://github.com/acmesh-official/acme.sh/issues/4659) 2 days ago, already exploited by the (former) chinese CA 'HiCA' (The issue is **very** entertaining to read btw :smirk:).

To be sure I've executed:

```console
$ docker exec $container-name cat /app/acme.sh | grep "VER="
VER=2.9.0
```

I have not tested if a RCE is possible though.

**Solution:**

1. Check if acme.sh can be updated to the latest version (hotfix, v3.0.6)
    - Shouldn't cause problems. Only v3.0.0 looks like a bigger change - But verify by yourslef.
2. Replace version in the [Dockerfile#L6](https://github.com/nginx-proxy/acme-companion/blob/63176c28eaeeb76dfb2f2e69f9fc03f045b3d206/Dockerfile#L6) to download the newer script

That _should_ be all, but I don't know since I'm not involved in this project.

acme-companion image version
-----------------

```
Info: running acme-companion version v2.2.8
```

nginx-proxy's Docker configuration, rendered nginx configuration, Containers logs, Docker host
-----------------

N/A

buchdag:
Redo of #1063 with fixes.

Fixes #1035

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,11 +12,18 @@ ENV COMPANION_VERSION=$GIT_DESCRIBE \
 # Install packages required by the image
 RUN apk add --no-cache --virtual .bin-deps \
     bash \
+    bind-tools \
     coreutils \
     curl \
     jq \
+    libidn \
+    oath-toolkit-oathtool \
+    openssh-client \
     openssl \
-    socat
+    sed \
+    socat \
+    tar \
+    tzdata
 
 # Install docker-gen from the nginxproxy/docker-gen image
 COPY --from=docker-gen /usr/local/bin/docker-gen /usr/local/bin/
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ FROM nginxproxy/docker-gen:0.11.0 AS docker-gen
 FROM alpine:3.19.0
 
 ARG GIT_DESCRIBE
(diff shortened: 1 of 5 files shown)


buchdag:
@pini-gh could you take a look at this PR ?

Also, do you think we should update the dependencies installed inside the image to match those in [acme.sh's Dockerfile](https://github.com/acmesh-official/acme.sh/blob/3.0.7/Dockerfile#L4-L15) ?

I guess some of those are needed for some of the DNS integrations, right ?

pini-gh:
> @pini-gh could you take a look at this PR ?

LGTM.

> Also, do you think we should update the dependencies installed inside the image to match those in acme.sh's Dockerfile ?
>
> I guess some of those are needed for some of the DNS integrations, right ?

Yes. We have to trust acme.sh to know its dependencies well.

buchdag:
I added acme.sh missing dependencies except `cronie`, we really don't want a cron process running inside the container.

PR #1092, opened 2024-01-28
fix: get container ID from /proc/self/mountinfo (for podman)

rioriost:
With cgroup v2, CID can be read from /proc/self/mountinfo

OCI specification defines that hostname must be set with UTS namespace.
http://wking.github.io/opencontainer-runtime-spec/#_hostname

Podman set the hostname of container running in a pod to a name of pod, e.g. 'test-pod', not a container ID, e.g. '1cb364360e42'.
Docker API via /tmp/docker.sock linked to /var/run/podman.sock returns error as following.
{"cause":"no such container","message":"\"test-pod\" is a pod, not a container: no such container","response":404}

Need to get container ID from /proc/self/mountinfo from the container inside if the container is running in a pod by Podman.

### DIFF
--- a/app/functions.sh
+++ b/app/functions.sh
@@ -199,14 +199,14 @@ function get_self_cid {
 
     # Try the /proc files methods first then resort to the Docker API.
     if [[ -f /proc/1/cpuset ]]; then
-        self_cid="$(grep -Eo '[[:alnum:]]{64}' /proc/1/cpuset)"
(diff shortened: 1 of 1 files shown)


buchdag (approved):

buchdag:
Hi.

Does neither the `/proc/1/cpuset` nor `/proc/self/cgroup` methods work on Podman ?

buchdag:
friendly ping @rioriost

buchdag (dismissed):

buchdag (approved):

PR #1128, opened 2024-06-10
feat: standalone certs with LETSENCRYPT_HOST

pini-gh:
This commit enables the creation of standalone certificates using the environment variable `LETSENCRYPT_HOST` without using `VIRTUAL_HOST`.

@buchdag would you be open to such an idea?

### DIFF
--- a/app/functions.sh
+++ b/app/functions.sh
@@ -141,7 +141,9 @@ function add_standalone_configuration {
     local domain="${1:?}"
     if grep -q "server_name ${domain};" /etc/nginx/conf.d/*.conf; then
         # If the domain is already present in nginx's conf, use the location configuration.
-        add_location_configuration "$domain"
+        if parse_true "${ACME_HTTP_CHALLENGE_LOCATION:=false}"; then
+            add_location_configuration "$domain"
+        fi
     else
         # Else use the standalone configuration.
         cat > "/etc/nginx/conf.d/standalone-cert-$domain.conf" << EOF
--- a/app/letsencrypt_service
+++ b/app/letsencrypt_service
@@ -349,9 +349,7 @@ function update_cert {
         # Add all the domains to certificate
         params_issue_arr+=(--domain "$domain")
         # If enabled, add location configuration for the domain
-        if parse_true "${ACME_HTTP_CHALLENGE_LOCATION:=false}"; then
-            add_location_configuration "$domain" || reload_nginx
-        fi
+        add_standalone_configuration "$domain" || reload_nginx
     done
 
     params_issue_arr=("${params_base_arr[@]}" "${params_issue_arr[@]}")


buchdag:
@pini-gh I don't really get what this PR intend to do 🤔

Unless I'm mistaken, it undoes https://github.com/nginx-proxy/acme-companion/pull/1123, which should trigger https://github.com/nginx-proxy/nginx-proxy/issues/2469 again when using both `LETSENCRYPT_HOST` and `VIRTUAL_HOST`. I don't really understand why / how some tests are passing with this change.

pini-gh:
The intent is to create standalone certificates with `LETSENCRYPT_HOST` for services with no HTTP exposure (no `VIRTUAL_HOST` variable). For instance an SMTP server: 
```
version: '3.5'

services:
  smtp:
    image: mysmtp:latest
    restart: unless-stopped
    environment:
      LETSENCRYPT_HOST: smtp.example.org
```

Without this change we have to use the file `letsencrypt_user_data` for that because there is no nginx-proxy configuration generated for this container. But since our certificate can be tied to a container we could just use `LETSENCRYPT_HOST` which is easier.

buchdag:
@pini-gh while I do get the idea and the easier / more user friendly approach, I'm not at ease with the shift in project scope that change would introduce by removing the clear separation between the original scope of the project (obtaining ACME SSL certificate for containers proxied by nginx-proxy) and a secondary feature added by request (obtaining SSL certificate not tied to specific containers, ie standalone certificate).

pini-gh:
I fail to see it as a shift because the feature exists already with `letsencrypt_user_data`. This change just makes it easier and more consistent.

For a web app you often need to glue a few containers. together into a docker-compose file Some of them with no HTTP may need certificates as well (SMTP, IMAP, ...). When looking at the related docker-compose file it's weird that one container can have its certificate with `LETENCRYPT_HOST` and the other cannot just because it has no HTTP. Moreover you have to tinker into another docker-compose file (acme-companion's one) to obtain this certificate.

buchdag:
If we set aside the project scope discussion which is kinda debatable, I'm not opposed to this in theory as it make sense and would be a nice usability improvement.

What concern me with the concrete implementation is the fact that now every certificate we request trigger this function:
```bash
function add_standalone_configuration {
    local domain="${1:?}"
    if grep -q "server_name ${domain};" /etc/nginx/conf.d/*.conf; then
        # If the domain is already present in nginx's conf, use the location configuration.
        if parse_true "${ACME_HTTP_CHALLENGE_LOCATION:=false}"; then
            add_location_configuration "$domain"
        fi
    else
        # Else use the standalone configuration.
        cat > "/etc/nginx/conf.d/standalone-cert-$domain.conf" << EOF
server {
    server_name $domain;
    listen 80;
    access_log /var/log/nginx/access.log vhost;
    location ^~ /.well-known/acme-challenge/ {
        auth_basic off;
        auth_request off;
        allow all;
        root /usr/share/nginx/html;
        try_files \$uri =404;
        break;
    }
}
EOF
    fi
}
```

Doesn't that mean that everyone who runs acme-companion without sharing the `/etc/nginx/conf.d` folder between nginx-proxy and acme-companion (ie the normal two containers setup), would have its logs filled with the following for every domain (because in this setup `/etc/nginx/conf.d` does not exist inside acme-companion container) ?
```console
grep: /etc/nginx/conf.d/*.conf: No such file or directory
bash: /etc/nginx/conf.d/standalone-cert-test.domain.tld.conf: No such file or directory
```

Additionally when I originally wrote this code I was okay with _"if we grep the domain in the rendered nginx conf generate a `location` block, otherwise generate a `server` block"_ being used for standalone certificates, but I'm not confortable with the "main use case" certificate (ie certificates tied to a container proxied through nginx-proxy) triggering this. I'm concerned that we could mistakenly go into the `else` branch of `if grep` because somehow this function was called before the nginx config was fully rendered, generate a `server` block that we shouldn't, and break the nginx config.

I think it might be possible to identify which `LETSENCRYPT_HOST` are tied to containers with `VIRTUAL_HOST` / `VIRTUAL_HOST_MULTIPORTS` with a modification to `letsencrypt_service_data.tmpl`.

pini-gh:
Your concerns about the implementation are valid. The current status of this PR is POC and it needs some work.

buchdag:
> would you be open to such an idea?

That's a yes then, let me know if you need help to refine this.

PR #1149, opened 2024-09-11
feat: specific error for socket permission

JohnMertz:
As discussed here: https://github.com/nginx-proxy/acme-companion/issues/1054

If socket is not readable, for example when protected with SELinux and/or run as a rootless user, the existing error is deceptive. An additional error to check that it is readable before checking if it is a socket will clarify the problem. SELinux tip included for good measure.


buchdag (approved):

buchdag:
Thanks @JohnMertz

PR #1155, opened 2024-09-27
Docs: Clarify standalone dns01 API configuration structure with example

stuartbirrell:
As discussed. Added a little more detail into the standalone configuration doc to be more specific around what is required when defining a specific DNS API configuration for a standalone cert.


buchdag (approved):

buchdag:
Thank you @stuartbirrell 👍

PR #1156, opened 2024-10-01
docs: update Basic-usage.md

yoriiis:
There is a difference between docker compose example and docker run for the acme-companion service about the certificat volume.

https://github.com/nginx-proxy/acme-companion/blob/main/docs/Docker-Compose.md?plain=1#L45

I imagine that the docker compose is the right syntax

### DIFF
--- a/docs/Basic-usage.md
+++ b/docs/Basic-usage.md
@@ -37,6 +37,7 @@ $ docker run --detach \
     --name nginx-proxy-acme \
     --volumes-from nginx-proxy \
     --volume /var/run/docker.sock:/var/run/docker.sock:ro \
+    --volume certs:/etc/nginx/certs:rw \
     --volume acme:/etc/acme.sh \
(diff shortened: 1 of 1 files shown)


buchdag:
Hi.

In the compose file example, the nginx-proxy container mount the certs volume as `:ro`, so it need to be explicitly mounted as `:rw` on the acme-companion container.

In the Docker run example, the command for the nginx-proxy container does not specify `:ro` or `:rw` for this volume, so it defaults to `:rw` and the acme-companion container get the volume with the correct read/write permission through `--volumes-from nginx-prox` alone.

Both versions will work equally well, it boils down to restricting nginx-proxy write permission to the certs volume or not.

yoriiis:
Hello @buchdag thanks for the explanation.

It might be a good idea to harmonize/synchronize to have the same implementation, but I'll let you decide that.

Thanks, you're free to close the issue.

PR #1157, opened 2024-10-04
feat: disable self signed default certificate creation

buchdag:
This PR disable the self signed default certificate creation feature.

It can be re-enabled by setting `CREATE_DEFAULT_CERTIFICATE` to `true` on the acme-companion container.

[This feature was added](https://github.com/nginx-proxy/acme-companion/pull/423) at a time when nginx-proxy had a bug that made it incorrectly serve (ie leak) the certificate for other domain(s) than the one that was request.

This is no longer the case, and forcing the creation of a self signed default certificate is now counter productive, with some users not understanding how they end up with a strange self signed certificate.

ping @pini-gh I think we discussed disabling this a few months ago but can't remember where.


pini-gh:
> ping @pini-gh I think we discussed disabling this a few months ago but can't remember where.

I don't remember either. But I agree that having a self signed default certificate does not help.

Issue #1179, opened 2025-01-06
Error: Docker host socket at /var/run/docker.sock is not readable


andrei-datcu:
```
REPOSITORY                  TAG                 IMAGE ID            CREATED             SIZE
nginxproxy/acme-companion   latest              a4f6c6c20efa        3 weeks ago         46.6MB
```
-----------------

```console
Info: running acme-companion version v2.5.1-6-gea11f22
Error: Docker host socket at /var/run/docker.sock is not readable. Please check user permissions
If you are in a SELinux environment, try using: '-v /var/run/docker.sock:/var/run/docker.sock:z'

```

Note that the socket seems to be readable. Binding with `:z` doesn't make any difference:

Proof:
```console
348d706f16fe:/app# bash ./entrypoint.sh /bin/bash /app/start.sh
Info: running acme-companion version v2.5.1-6-gea11f22
Error: Docker host socket at /var/run/docker.sock is not readable. Please check user permissions
If you are in a SELinux environment, try using: '-v /var/run/docker.sock:/var/run/docker.sock:z'
348d706f16fe:/app# ls -alh /var/run/docker.sock 
srw-rw---- 1 root 998 0 Aug 12 13:24 /var/run/docker.sock
348d706f16fe:/app# 
exit
```

Executing `start.sh` **works**.

Docker host
-----------------

 - OS: `Debian 10 Linux debian 4.19.0-27-amd64 #1 SMP Debian 4.19.316-1 (2024-06-25) x86_64 GNU/Linux`
 - Docker version:
```Client: Docker Engine - Community
 Version:           19.03.8
 API version:       1.40
 Go version:        go1.12.17
 Git commit:        afacb8b7f0
 Built:             Wed Mar 11 01:25:56 2020
 OS/Arch:           linux/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          19.03.8
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.12.17
  Git commit:       afacb8b7f0
  Built:            Wed Mar 11 01:24:28 2020
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.2.13
  GitCommit:        7ad184331fa3e55e52b890ea95e65ba581ae3429
 runc:
  Version:          1.0.0-rc10
  GitCommit:        dc9208a3303feef5b3839f4323d9beb36df0a9dd
 docker-init:
  Version:          0.18.0
[...]

PR #1182, opened 2025-01-18
fix: reverse docker socket test

buchdag:
Fix proposed by @andrei-datcu for #1179 

The entrypoint will fail if `test -S $socket_file` is `false` but only display a warning if `test -r $socket_file` is `false`.

### DIFF
--- a/app/entrypoint.sh
+++ b/app/entrypoint.sh
@@ -14,14 +14,14 @@ function print_version {
 function check_docker_socket {
     if [[ $DOCKER_HOST == unix://* ]]; then
         socket_file=${DOCKER_HOST#unix://}
-	if [[ ! -S $socket_file ]]; then
+        if [[ ! -S $socket_file ]]; then
+            if [[ ! -r $socket_file ]]; then
+                echo "Warning: Docker host socket at $socket_file might not be readable. Please check user permissions" >&2
+                echo "If you are in a SELinux environment, try using: '-v /var/run/docker.sock:$socket_file:z'" >&2
+            fi
             echo "Error: you need to share your Docker host socket with a volume at $socket_file" >&2
             echo "Typically you should run your container with: '-v /var/run/docker.sock:$socket_file:ro'" >&2
             exit 1
-        fi
-        if [[ ! -r $socket_file ]]; then
-            echo "Warning: Docker host socket at $socket_file might not be readable. Please check user permissions" >&2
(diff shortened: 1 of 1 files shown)


app/entrypoint.sh
```diff
@@ -14,16 +14,15 @@ function print_version {
 function check_docker_socket {
     if [[ $DOCKER_HOST == unix://* ]]; then
         socket_file=${DOCKER_HOST#unix://}
-	if [[ ! -r $socket_file ]]; then
-            echo "Error: Docker host socket at $socket_file is not readable. Please check user permissions" >&2
-            echo "If you are in a SELinux environment, try using: '-v /var/run/docker.sock:$socket_file:z'" >&2
-            exit 1
-        fi
-        if [[ ! -S $socket_file ]]; then
+	    if [[ ! -S $socket_file ]]; then
```
buchdag, line 17:
```suggestion
	if [[ ! -S $socket_file ]]; then
```

andrei-datcu:
I don't have the proper historical context of why the `-r` check was introduced, but for the original problem the check tried to resolve would this change work or would the `-S` check fail before `-r` test would get a chance to run? I was thinking the `-r` check would only be done if the `-S` failed to try and output a more explicit error message.

buchdag:
The `-r` check was added here https://github.com/nginx-proxy/acme-companion/pull/1149

So maybe something like this instead ?

```bash
if [[ ! -S $socket_file ]]; then
    if [[ ! -r $socket_file ]]; then
        echo "Warning: Docker host socket at $socket_file might not be readable. Please check user permissions" >&2
        echo "If you are in a SELinux environment, try using: '-v /var/run/docker.sock:$socket_file:z'" >&2
    fi
    echo "Error: you need to share your Docker host socket with a volume at $socket_file" >&2
    echo "Typically you should run your container with: '-v /var/run/docker.sock:$socket_file:ro'" >&2
    exit 1
fi

```

andrei-datcu:
Yes, that looks better to me.

buchdag:
Done, pushed as [`nginxproxy/acme-companion:1182`](https://hub.docker.com/repository/docker/nginxproxy/acme-companion/tags/1182/sha256-573af5e51d65ec326c14e427ed18e7bef067dafd009597b91055ab09a4148948) if you wish to try it before we merge this.

andrei-datcu:
I've just tested it and it's working nicely. Thank you! :)

PR #1189, opened 2025-04-01
Remove version and default bridge network

emfox:
1. 'version' is obsolete.
2. Connect to default bridge is not recommended, just let docker create a 'nginx-proxy_default' bridge by default, or define named network.


emfox:
"The default bridge network is considered a legacy detail of Docker and is not recommended for production use. " from official documentation:
https://docs.docker.com/engine/network/drivers/bridge/#use-the-default-bridge-network

buchdag:
Thanks @emfox 👍

PR #1205, opened 2025-08-12
Fix doc: add html volume to acme-companion service

raulodev:
The acme-companion service needs access to the HTML volume where the Let's Encrypt validation file (.well-known/acme-challenge/...) is located.


buchdag:
As stated in the doc this example is for compose files specification v2 exclusively, as `volumes_from` was later removed from compose file specification.

Maybe the compose examples should be re written with a more up to date compose file specification in mind 🤷🏻‍♂️

PR #1211, opened 2025-09-17
feat: include stopped containers for creation/renewal

ftsartek:
The intent here is to start a discussion on the viability of enabling certificate renewal for stopped containers; useful when you might want NGINX proxy to serve an 'unavailable' page or similar with a valid certificate.

Seemingly a simple change, though might need some test cases. Would you be open to including this or something similar?

### DIFF
--- a/app/start.sh
+++ b/app/start.sh
@@ -20,7 +20,11 @@ letsencrypt_service_pid=$!
 
 wait_default="5s:20s"
 DOCKER_GEN_WAIT="${DOCKER_GEN_WAIT:-$wait_default}"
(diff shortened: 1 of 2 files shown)


buchdag:
@ftsartek this will be indirectly made possible by an upstream feature on docker-gen

See https://github.com/nginx-proxy/docker-gen/pull/623 and https://github.com/nginx-proxy/docker-gen/pull/700, once the later is merged (and a docker-gen version released then used in this project, etc) you'll just have to correctly set the `DOCKER_CONTAINER_FILTERS` environment variable on your acme-companion container for this to work.

ftsartek:
Ah, that's awesome. Glad to hear it, so this is effectively redundant. Thank you!

PR #1230, opened 2025-12-18
feat: add support for certificate profiles

brakhane:

PR #1256, opened 2026-05-19
feat: per-container certificate renewal timing

buchdag:
This PR
- rename the global environment variable `DEFAULT_RENEW` to `ACME_RENEW_AFTER` to clarify what it does (while maintaining backward compatibility with the old variable for the time being)
- add support for `ACME_RENEW_AFTER` at the application container level in addition to the global acme-companion container level

@brakhane this feature was motivated by the introduction of certificate profiles support in #1230 , so if you have some time to spare I'd like to have your feedback on this 👍

### DIFF
--- a/docs/Let's-Encrypt-and-ACME.md
+++ b/docs/Let's-Encrypt-and-ACME.md
@@ -125,7 +125,7 @@ If the ACME CA provides multiple cert chain, you can use the `ACME_PREFERRED_CHA
 
 #### Certificate profile
 
-The `ACME_CERT_PROFILE` environment variable is used to select a specific profile offered by the CA. See for example [the list of profiles offered by Letsencrypt](https://letsencrypt.org/docs/profiles). Note that some profiles might reduce the validity period of the certificate; you might need to adjust the (global) `DEFAULT_RENEW` variable to make sure it gets updated in time.
+The `ACME_CERT_PROFILE` environment variable is used to select a specific profile offered by the CA. See for example [the list of profiles offered by Letsencrypt](https://letsencrypt.org/docs/profiles). Note that some profiles might reduce the validity period of the certificate; you might need to adjust the (global) `ACME_RENEW_AFTER` variable or set it per-container to make sure it gets updated in time.
 
 #### Container restart on cert renewal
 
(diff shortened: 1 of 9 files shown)


docs/Let's-Encrypt-and-ACME.md
```diff
@@ -135,6 +135,10 @@ The `LETSENCRYPT_RESTART_CONTAINER` environment variable, when set to `true` on
 
 The `ACME_PRE_HOOK` and `ACME_POST_HOOK` let you use the [`acme.sh` Pre- and Post-Hooks feature](https://github.com/acmesh-official/acme.sh/wiki/Using-pre-hook-post-hook-renew-hook-reloadcmd) to run commands respectively before and after the container's certificate has been issued. For more information see [Pre- and Post-Hook](./Hooks.md)
 
+#### Certificate renewal timing
+
+The `ACME_RENEW_AFTER` environment variable can be set on an application container to override the global renewal timing for that specific container's certificate. This is useful for certificates from CAs with different validity periods. For example, Buypass certificates have a 180-day lifespan, so you might want to set `ACME_RENEW_AFTER=150` on those containers while keeping the default 60 days for Let's Encrypt certificates on others.
```
brakhane: Since it's already mentioning letsencrypt, it might be good to also mention `shortlived` certificates here

buchdag: Proposed change:

> This is useful for certificates from CAs with different validity periods, or when using specific [certificate profiles](#certificate-profile), such as Let's Encrypt [`shortlived`](https://letsencrypt.org/docs/profiles/#shortlived) profile.

brakhane: A certificate profile does not necessarily have a different validity, how about something like this to make it clearer:

> This is useful when using a CA or a [certificate profile](#certificate-profile) with a different validity period; for example, Buypass certificates have a lifespan of 180 days, and Let's Encrypt's offers [`tlsserver`](https://letsencrypt.org/docs/profiles/#tlsserver) (45 days) and [`shortlived`](https://letsencrypt.org/docs/profiles/#shortlived)  (180 hours) profiles.

I'm not sure if having `tlsserver` as another example helps with clarity or makes it worse. I've added it to show lifetimes can vary widely from one profile to the next, but it might be Too Much Information™.

(NB: Github seems to mangle the anchor link for certificate profiles and turns it into a link to this PR)

buchdag: That seems fine to me, and well who read the docs anyway 😅

brakhane (commented):
See my comment above (hope this makes GitHub happy)

docs/Let's-Encrypt-and-ACME.md
brakhane: > who read the docs anyway

Claude and ChatGPT ;)

PR #1267, opened 2026-06-18
feat: IPv6 support for standalone challenge config

### LINKED ISSUE #710: Unable to generate standalone certs with ipv6 enabled
Hello,

I was trying to generate a cert for a standalone service but it wasn't working for some reasons. After a little digging, I found that trying to reach `http://example.com/.well-known/acme-challenge/` was not working during the cert generation loop (after a manual call to `/app/signal_le_service`) even if the file `standalone-cert-example.com.conf` was in `/etc/nginx/conf.d/`. I tried to replicate the issue with a copy of the previously generated file and found out that it was missing the line `listen [::]:80;`(maybe on purpose ?). I'm currently on a dual stack server and if I correctly understood  the error message I got from the companion[1], Let's Encrypt will perform the challenge validation over IPv6 if my domain's DNS entry has both A and AAAA fields. I added the proper line and everything worked fine.

I don't know if it's a good idea to add the line `listen [::]:80;` by default in the `add_standalone_configuration function` ? Or maybe we could find a way to add it dynamically ?

https://github.com/nginx-proxy/docker-letsencrypt-nginx-proxy-companion/blob/711c56acc160ea921cbc5797b5dc39d28df044bc/app/functions.sh#L115

[1] CA marked some of the authorizations as invalid, which likely means it could not access http://example.com/.well-known/acme-challenge/X. Did you set correct path in -d example.com:path or --default_root? Are all your domains accessible from the internet? Please check your domains' DNS entries, your host's network/firewall setup and your webserver config. If a domain's DNS entry has both A and AAAA fields set up, some CAs such as Let's Encrypt will perform the challenge validation over IPv6. If your DNS provider does not answer correctly to CAA records request, Let's Encrypt won't issue a certificate for your domain (see https://letsencrypt.org/docs/caa/). Failing authorizations: https://acme-v02.api.letsencrypt.org/acme/authz-v3/8466860020

Thanks,

Byh0ki

JamBalaya56562:
## What

Add an `ENABLE_IPV6` toggle that makes the **standalone** ACME HTTP-01 challenge configuration also listen over IPv6 (`listen [::]:80;`), in addition to IPv4.

## Why

Standalone challenge configs are generated by acme-companion's `add_standalone_configuration` (not by docker-gen), and they only emitted `listen 80;`. On IPv6-only / dual-stack hosts the HTTP-01 challenge was therefore unreachable over IPv6, so validation failed.

This revives #710 (open since 2020), which @buchdag endorsed in-thread.

## How

- **`app/functions.sh`** — `add_standalone_configuration` appends `listen [::]:80;` when `ENABLE_IPV6` is truthy (gated via the existing `parse_true` helper). It is **opt-in**, so hosts without IPv6 are unaffected (nginx won't try to bind an IPv6 socket) — this addresses @buchdag's concern about enabling it by default.
- The variable name matches **nginx-proxy**'s own [`ENABLE_IPV6`](https://github.com/nginx-proxy/nginx-proxy#ipv6-support), so the same value can be set on both containers. (Of the two options discussed in the issue — auto-detecting nginx-proxy's setting vs. our own env var — this implements the env-var option, which is self-contained and deterministic.)
- **`docs/Container-configuration.md`** — documents `ENABLE_IPV6`.
- **`test/tests/standalone_ipv6/`** — deterministic test asserting the IPv6 listener is present only when `ENABLE_IPV6` is enabled; registered in `test/config.sh` and the CI matrix.

Only the standalone config is affected; challenges served through nginx-proxy already follow nginx-proxy's own IPv6 setting.

## Testing

- New `standalone_ipv6` test passes: asserts `listen [::]:80;` is present with `ENABLE_IPV6=true` and absent when `false`/unset.
- `shellcheck` clean on the modified script and the new test.

Closes #710


app/functions.sh
```diff
@@ -135,10 +135,15 @@ function add_standalone_configuration {
         add_location_configuration "$domain"
     else
         # Else use the standalone configuration.
+        local listen_directives='    listen 80;'
+        # Also listen over IPv6 when ENABLE_IPV6 is set (same toggle as nginx-proxy). See #710.
```
buchdag: Again, needlesly verbose AI comment, please remove.

JamBalaya56562: Done, removed in 6c9aef2 — thanks!

buchdag (approved):

PR #1269, opened 2026-06-18
test: fix flaky force_renew test (detect renewal by serial)

JamBalaya56562:
## What

Make the `force_renew` integration test reliable. It has been failing intermittently in CI (maintainers have had to re-run it to get a green build).

## Why

The test forced a renewal and then polled for the certificate's **expiration date** (`notAfter`) to increase. That signal is unreliable: `test/setup/pebble/pebble-config.json` defines several certificate profiles with different validity periods (5y default, 1y, 6-day), and a renewed certificate can be issued under a *different* profile than the original — giving it an **earlier** `notAfter` even though the renewal succeeded. The poll then never sees `notAfter` increase and the test fails after 30s.

Example from a failed run:

```
First  certificate expiration epoch : 1939542803   # issued + ~5y  (default profile, 157766400s)
Second certificate expiration epoch : 1813312416   # issued + ~1y  (long-lived profile, 31536000s)
```

The two issuances were only ~13s apart (so the renewal *did* happen), but the renewed cert got a shorter validity, so `notAfter` went backwards.

## How

Detect the renewal by the certificate **serial number**, which changes on every (re-)issuance regardless of profile, validity period or timing:

- `test/tests/test-functions.sh` — add a `get_cert_serial` helper.
- `test/tests/force_renew/run.sh` — compare the serial before/after the forced renewal instead of the expiration date, and capture the `force_renew` output so a genuine failure is diagnosable (it was previously discarded to `/dev/null`).

Test-only change; no product code is affected.

### DIFF
--- a/test/tests/force_renew/run.sh
+++ b/test/tests/force_renew/run.sh
@@ -36,8 +36,7 @@ sleep 5
 # Issue a forced renewal (capture the output so a failure is diagnosable).
 renew_output="$(docker exec "$le_container_name" /app/force_renew 2>&1)"
 
-# A renewal re-issues the cert, so its serial must change. We check the serial rather
-# than the expiration date, which is unreliable here (the Pebble test profiles differ).
+# A renewal re-issues the cert, so its serial must change.
 timeout=$(($(date +%s) + 30))
 second_serial="$first_serial"
 while [[ $(date +%s) -lt $timeout ]]; do
--- a/test/tests/force_renew/run.sh
+++ b/test/tests/force_renew/run.sh
@@ -26,42 +26,33 @@ trap cleanup EXIT
 # Run a nginx container for ${domains[0]}.
 run_nginx_container --hosts "${domains[0]}"
 
-# Wait for a symlink at /etc/nginx/certs/${domains[0]}.crt
-# Grab the expiration time of the certificate
+# Wait for the certificate to be issued, then record its serial number.
 wait_for_symlink "${domains[0]}" "$le_container_name"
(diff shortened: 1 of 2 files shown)


test/tests/test-functions.sh
```diff
@@ -348,6 +348,17 @@ function get_cert_date_epoch {
 }
 export -f get_cert_date_epoch
 
+# Get the serial number of the certificate for domain $1 inside container $2.
+# The serial changes on every (re-)issuance, so it is a reliable signal that a
+# certificate was renewed (unlike the expiration date, which depends on the CA's
+# certificate profile / validity period).
```
buchdag: Bla bla bla you get the picture 😆 
```suggestion
# Get the serial number of the certificate for domain $1 inside container $2.
```

test/tests/force_renew/run.sh
```diff
@@ -26,42 +26,33 @@ trap cleanup EXIT
...
 
-# Issue a forced renewal
-docker exec "$le_container_name" /app/force_renew &> /dev/null
+# Issue a forced renewal (capture the output so a failure is diagnosable).
+renew_output="$(docker exec "$le_container_name" /app/force_renew 2>&1)"
 
-# Poll until expiration date changes or timeout
-# Use a longer sleep and add error handling for transient states
+# A renewal re-issues the cert, so its serial must change. We check the serial rather
+# than the expiration date, which is unreliable here (the Pebble test profiles differ).
```
buchdag:
```suggestion
# A renewal re-issues the cert, so its serial must change.
```

buchdag (approved):

PR #1272, opened 2026-06-19
docs: modernize the Docker Compose examples

### LINKED ISSUE #664: Doc updates for V3 Docker compose?
There are a lot of wildly varying examples on the web for using this with Docker compose v3, are there any thoughts about updating the docs to support it?

JamBalaya56562:
## What

Modernize the Docker Compose examples in `docs/Docker-Compose.md`. Both the two and three containers examples used the obsolete top-level `version:` key and `volumes_from` (a compose file version 2 only feature), so they could not be copy-pasted into a current Compose Specification file.

## Changes

- Drop the `version:` key — it is obsolete in the [Compose Specification](https://docs.docker.com/reference/compose-file/) and ignored by Docker Compose.
- Replace `volumes_from` with explicit per-container volume mounts.
- Detect the nginx/nginx-proxy and docker-gen containers with the `label` method (`com.github.nginx-proxy.nginx` / `com.github.nginx-proxy.docker-gen`). This also replaces the `volumes_from` container-detection fallback that is being removed.
- Mount the shared `html` volume read-write on acme-companion so it can write the HTTP-01 challenge files that nginx serves.

Both examples validate with `docker compose config`.

## Scope

This PR only touches `docs/Docker-Compose.md`. The `docker run --volumes-from` method documented in Basic-usage / Advanced-usage / Getting-containers-IDs remains valid and is intentionally left unchanged.

Closes #664

### DIFF
--- a/docs/Docker-Compose.md
+++ b/docs/Docker-Compose.md
@@ -6,11 +6,15 @@ As stated by its repository, [Docker Compose](https://github.com/docker/compose)
 
 Be sure to be familiar with both the [basic](./Basic-usage.md) and [advanced](./Advanced-usage.md) non compose setups, and Docker Compose usage.
 
-Please read [getting container IDs](./Getting-containers-IDs.md) and be aware that the `--volumes-from method` is **only** available on compose file version 2.
+The following examples are minimal, clean starting points, not a definitive reference. They follow current Docker Compose conventions:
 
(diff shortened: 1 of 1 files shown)


buchdag (approved):

buchdag:
@JamBalaya56562 please keep the PR titles and commit message size reasonably small.

image

PR #1279, opened 2026-07-06
test: fix renew_private_keys repeated failure

buchdag:
This PR attempts to fix the `renew_private_keys` test that has been failing repeatedly since #1274 by reproducing the solution implemented in `test/tests/force_renew/run.sh` (ie checking that the certificate serial changes rather than checking that the expiration timestamp changes).


buchdag:
I ran the fixed test a few times in a row with success every time.

